PR #3, opened 2021-08-18
Dunitz/data adaptor metadata api

MDunitz:
#### Reviewers
**Functional:** 
@bkmartinjr 
**Readability:** 
@maniarathi 
---

## Changes

This is setting up hosted cellxgene to use  /datasets/meta?url={} in the portal to retrieve information (collection id, collection visibility, s3 uri, dataset id and tombstone status) about the dataset based on the url) and then pull the dataset based on the location given in the s3_uri.
For further detail see [ticket 2234](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/2234) and [RFC](https://github.com/chanzuckerberg/single-cell/blob/main/design-documents/0006-metadata-api/text.md)

## Approach
Split apart caching and dataset management. 
Add an additional cache to handle dataset location (redirects from explorer location).  

This is a continuation of a pr started in the cellxgene repo. To see the original comments/discussion on the pr see https://github.com/chanzuckerberg/cellxgene/pull/2298


backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+                try:
+                    self.data.cleanup()
+                except AttributeError:
+                    pass
+                self.data = None
+
+    def attempt_delete(self):
+        """Delete, but only if the write lock can be immediately locked.  Return True if the delete happened"""
+        if self.data_lock.w_acquire_non_blocking():
+            if self.data:
```
maniarathi, line 67: Just an edge case here.. if self.data is None, then this function would return None. Is that expected/handled?

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+        self.data = {}
+
+        # lock to protect the datasets
+        self.lock = threading.Lock()
+
+        #  The number of items to cache.  When max_cached is reached, the least recently used
+        #  cache is replaced with the newly requested one.
+        #  TODO:  This is very simple.  When this class is used to cache the actual datasets
+        #   this can be improved by taking into account how much space is actually taken by each dataset,
+        #   instead of arbitrarily picking a max number to cache.
```
maniarathi, line 131: Nit: sorry this is _really_ a tiny nit. The left most edge of the comments are not aligned by 1 space :P

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+            return False
+
+
+class CacheItemInfo(object):
+    """
+    This class stores a pointer to and metadata about the cached item. When the data is accessd the
+    update function is called to track the latest access
+    """
+
+    def __init__(self, cache_item: CacheItem(), timestamp):
```
maniarathi, line 89:
```suggestion
    def __init__(self, cache_item: CacheItem(), timestamp: int):
```

Typing for consistency.

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+        self.max_cached = max_cached
+
+        # items are automatically removed from the cache once this time limit (in seconds) is reached
+        self.timelimit_s = timelimit_s
+
+    @contextmanager
+    def get(self, cache_key: str, create_data_function: Optional[Callable] = None,
+            create_data_args: object = {}):
+        """"
+        If no data is stored under the given cache_key, pass the create_data_lamba to the CacheItem.get method
```
maniarathi, line 141: Could you write a better pydoc here? I didn't quite get the gist of the function with the above comment. Also `create_data_lamba` doesn't exist? Should this be `create_data_function`? And why was it important to call out?

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+                del self.data[cache_key]
+                cache_item.delete()
+            cache_item = None
+            raise
+
+        finally:
+            if cache_item:
+                cache_item.release()
+
+    def evict_extra_data(self):
```
maniarathi, line 186: Add pydoc and note the LRU cache eviction policy.

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+            items = sorted(items, key=lambda x: x[1].last_access)
+            # close the least recently used loader
+            oldest = items[0]
+            oldest_cache = oldest[1].cache_item
+            oldest_key = oldest[0]
+            del self.data[oldest_key]
+            delete_adaptor = oldest_cache
+        return delete_adaptor
+
+    def evict_old_data(self):
```
maniarathi, line 202: Pydoc, with a note on the eviction policy.

backend/czi_hosted/data_common/dataset_metadata.py
```diff
@@ -0,0 +1,93 @@
...
+        "collection_id": None,
+        "collection_visibility": None,
+        "dataset_id": None,
+        "s3_uri": None,
+        "tombstoned": False
+    }
+    dataset_metadata["s3_uri"] = extrapolate_dataset_location_from_config(
+        server_config=server_config, dataset_explorer_location=dataset_explorer_location)
+    if dataset_metadata["s3_uri"] is None:
+        return common_rest.abort_and_log(HTTPStatus.BAD_REQUEST, "Invalid dataset NONE", loglevel=logging.INFO)
```
maniarathi, line 91: The error message here seems a bit convoluted -- could you make it a bit clearer please?

backend/czi_hosted/data_common/rwlock.py
```diff
@@ -29,7 +29,7 @@
 
 
 class RWLock(object):
-    """ RWLock class; this is meant to allow an object to be read from by
+    """ RWLock class; this is meant to allow an object to be read  by
```
maniarathi, line 32:
```suggestion
    """ RWLock class; this is meant to allow an object to be read by
```

backend/czi_hosted/data_cxg/cxg_adaptor.py
```diff
@@ -175,7 +175,6 @@ def _validate_and_initialize(self):
 
         if cxg_version not in ["0.0", "0.1", "0.2.0"]:
             raise DatasetAccessError(f"cxg matrix is not valid: {self.url}")
-
```
maniarathi, line 178: Think the extra line break here is nice so would revert changes in this file.

backend/czi_hosted/default_config.py
```diff
@@ -122,6 +122,15 @@
...
       timelimit_s: 30
 
+    metadata_cache:
+      # The maximum number of datasets that may be opened at one time.  The least recently used dataset
+      # is evicted from the cache first.
+      max_datasets: 10000
+
+      # A matrix is automatically removed from the cache after timelimit_s number of seconds.
+      # If timelimit_s is set to None, then there is no time limit.
+      timelimit_s: 300
```
maniarathi, line 132: Why is the time limit only 5 minutes? Seems like you would want longer?

maniarathi:
I just did a readability pass on this and it mostly looks fine -- just called out a few nits. I would rely on Bruce's review moving forward and use his LGTM to merge this.

bkmartinjr:
> I would rely on Bruce's review

I had already reviewed & approved when the PR lived in the cellxgene repo.  If it is the same, I'm good to go.

@MDunitz - LMK if you want another review (ie, there are changes)

backend/czi_hosted/data_common/cache.py
MDunitz: Based on where it is called that should be fine, but to be safe I'll release the lock and return True (to ensure the cache entry is deleted) when self.data is None

backend/czi_hosted/data_common/cache.py
MDunitz: 🤣

backend/czi_hosted/default_config.py
MDunitz: I was thinking that 5 minutes is long enough that loading a new dataset will only call the data portal once, but short enough that if we publish a revision the dataset location will be updated with minimal potential down time. 

But its super configurable/I'm open to changing it if you think a longer time period would be better. We could potentially add a delay longer than the cache length to the s3 object deletion when revising a dataset to avoid issues with downtime, but I'm a little worried about the user experience if a curator publishes a revision to a dataset and doesnt realize it will take a few minutes to propagate . What do you think?

maniarathi: Your logic makes sense to me! Happy to try it out and configure it based on the experience.

PR #4, opened 2021-08-19
TS typing of crossfilter

bkmartinjr:
This PR primarily adds TS typing for the crossfilter which is the last item in chanzuckerberg/cellxgene#2366
.  In addition:
* second pass through Dataframe and AnnoMatrix for additional cleanup (chanzuckerberg/cellxgene#2394)
* pulled typing through users of these API (eg, src/actions/annotations.ts)
* utils lint chanzuckerberg/cellxgene#2388

Fixes chanzuckerberg/cellxgene#2394
Fixes chanzuckerberg/cellxgene#2366
Fixes chanzuckerberg/cellxgene#2388

### DIFF
--- a/client/__tests__/util/typedCrossfilter/crossfilter.test.ts
+++ b/client/__tests__/util/typedCrossfilter/crossfilter.test.ts
@@ -1,7 +1,9 @@
 import filter from "lodash.filter";
 import zip from "lodash.zip";
 
-import Crossfilter from "../../../src/util/typedCrossfilter";
+import Crossfilter, {
+  CrossfilterSelector,
+} from "../../../src/util/typedCrossfilter";
 
 const someData = [
   {
@@ -126,8 +128,7 @@ const someData = [
   },
 ];
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-let payments: any = null;
+let payments: Crossfilter<typeof someData>;
 beforeEach(() => {
   payments = new Crossfilter(someData);
 });
@@ -139,13 +140,11 @@ describe("ImmutableTypedCrossfilter", () => {
     expect(payments.all()).toEqual(someData);
 
     const p = payments
-      .addDimension(
-        "quantity",
-        "scalar",
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-        (i: any, d: any) => d[i].quantity,
-        Int32Array
-      )
+      .addDimension("quantity", {
+        type: "scalar",
+        value: payments.all().map((d) => d.quantity),
+        ValueArrayCtor: Int32Array,
+      })
       .select("quantity", { mode: "all" });
     expect(p).toBeDefined();
     expect(p.all()).toEqual(someData);
@@ -162,13 +161,11 @@ describe("ImmutableTypedCrossfilter", () => {
     - delDimension()
     - select
     */
-    const p2 = payments.addDimension(
-      "quantity",
-      "scalar",
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-      (i: any, data: any) => data[i].quantity,
-      Int32Array
-    );
+    const p2 = payments.addDimension("quantity", {
+      type: "scalar",
+      value: payments.all().map((d) => d.quantity),
+      ValueArrayCtor: Int32Array,
+    });
 
     expect(payments).not.toBe(p2);
     const p3 = p2.select("quantity", { mode: "all" });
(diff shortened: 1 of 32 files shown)


client/__tests__/util/typedCrossfilter/crossfilter.test.ts
```diff
@@ -126,8 +128,7 @@ const someData = [
   },
 ];
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-let payments: any = null;
+let payments: Crossfilter<typeof someData>;
```
tihuan, line 131: Ooh never thought about using `typeof` this way! So convenient 🔥

tihuan (commented):
Still reviewing! But I just sent a PR to type some files a little more, in case other reviewers were thinking the same thing 😆 

https://github.com/chanzuckerberg/single-cell-explorer/pull/17

CC: @seve @MillenniumFalconMechanic 

Thank you!

client/src/util/typedCrossfilter/crossfilter.ts
```diff
@@ -8,32 +31,64 @@ import {
...
-    if (Error.captureStackTrace) {
-      Error.captureStackTrace(this, NotImplementedError);
-    }
-  }
+/**
+ * Table of dimensions added to the crossfilter
+ * @internal
+ */
+interface Dimension {
+  id: number | undefined; // ID is often undefined when selection cache is clear
```
tihuan, line 50:
```suggestion
  id?: number; // ID is often undefined when selection cache is clear
```

client/src/util/typedCrossfilter/crossfilter.ts
```diff
@@ -420,94 +498,75 @@ class _ImmutableBaseDimension {
...
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  value: any;
+  protected value: SortableArray;
 
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  constructor(name: any, data: any, value: any, ValueArrayType: any) {
+  // Two constructor modes - caller can provide a pre-created value array,
+  // a map function which will create it, or another array which
+  // will copied into the user-specified value array.
```
tihuan, line 507:
```suggestion
  // will be copied into the user-specified value array.
```

client/src/util/typedCrossfilter/crossfilter.ts
```diff
@@ -538,78 +596,66 @@ class ImmutableScalarDimension extends _ImmutableBaseDimension {
...
+  constructor(
+    name: string,
+    length: number,
+    params: SpatialDimensionParameters
+  ) {
     super(name);
 
-    if (X.length !== Y.length && X.length !== data.length) {
+    const { X, Y } = params;
+    if (X.length !== Y.length && X.length !== length) {
```
tihuan, line 658: would `||` catch more edge cases here?

client/src/util/typedCrossfilter/types.ts
```diff
@@ -0,0 +1,96 @@
...
+  SortableArray,
+  NumberArrayConstructor,
+  IntTypedArray,
+  UnsignedIntTypedArray,
+} from "../../common/types/arraytypes";
+
+/**
+ * Cross-filter shared types.
+ */
+
```
tihuan, line 12:
```suggestion

```

tihuan (approved):
LGTM!! Thanks again, Bruce! Mostly just tiny nits 👍

client/src/annoMatrix/annoMatrix.ts
```diff
@@ -468,22 +467,22 @@ export default abstract class AnnoMatrix {
...
     return queries
       .map((query: Query) =>
-        // @ts-expect-error --- TODO revisit:
-        // `filter`: This expression is not callable.
+        // @ts-expect-error ts-migrate --- suppressing TS defect (https://github.com/microsoft/TypeScript/issues/44373).
+        // Compiler is complaining that expression is not callable on array union types. Remove suppression once fixed.
         _whereCacheGet(this._whereCache, this.schema, field, query).filter(
-          (cacheKey?: LabelType) =>
+          (cacheKey: LabelType) =>
             cacheKey !== undefined && this._cache[field].hasCol(cacheKey)
```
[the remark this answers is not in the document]
bkmartinjr: No.  Please look at the definition of _whereCacheGet type signature.  An explicit `undefined` is returned and must be filtered out.

client/src/annoMatrix/annoMatrix.ts
```diff
@@ -502,7 +501,9 @@ Return cache keys for columns associated with this query.  May return
     /* find any query not already cached */
     const uncachedQueries = queries.filter((query) =>
       _whereCacheGet(this._whereCache, this.schema, field, query).some(
-        (cacheKey?: LabelType) =>
+        // @ts-expect-error ts-migrate --- suppressing TS defect (https://github.com/microsoft/TypeScript/issues/44373).
+        // Compiler is complaining that expression is not callable on array union types. Remove suppression once fixed.
+        (cacheKey: LabelType) =>
           cacheKey === undefined || !this._cache[field].hasCol(cacheKey)
```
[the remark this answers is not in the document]
bkmartinjr: No, this would introduce a bug!  See the type signature for _whereCacheGet.

client/src/util/typedCrossfilter/crossfilter.ts
bkmartinjr: No!  This is _not_ an optional field.  I want to enforce that it is always defined, even if the value is undefined.

I think you are confusing the (real) value `undefined` with the default behavior of object property access when a field is missing (eg, `foo.bar` evaluates to `undefined` when `bar` is missing).

client/src/util/typedCrossfilter/crossfilter.ts
bkmartinjr: will fix

client/src/util/typedCrossfilter/crossfilter.ts
bkmartinjr: yeah, that is a typo.   Will fix!

client/src/util/typedCrossfilter/types.ts
bkmartinjr: will fix.

client/src/util/typedCrossfilter/crossfilter.ts
tihuan: Sync'd with Bruce on this! The big upside is explicitness, where we require callsite to pass `undefined` for certain `key` instead of leaving them out as an optional argument:

E.g.,

```ts
function foo({ foo }: { foo: number | undefined }) { ... }

foo({}) // error
foo({ foo: undefined }) // good

function foo({ foo }: { foo?: number }) { ... }

foo({}) // good
foo({ foo: undefined }) // good
```

client/src/annoMatrix/annoMatrix.ts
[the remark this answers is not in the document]
tihuan: Ah sorry didn't think this through 🤔 

If `cacheKey` can be `undefined`, then `(cacheKey: LabelType) =>` should be `(cacheKey: LabelType | undefined) =>` ? Because `LabelType` is never `undefined`

And it seems like TS allows passing `undefined` to a function that accepts an optional argument:

```ts
function foo(bar?: number) {
  console.log(bar);
}

foo(0); // good
foo(); // good
foo(undefined); // good
foo(null); // bad
```
So `(cacheKey?: LabelType) =>` would work here too?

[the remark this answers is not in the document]
bkmartinjr: You are right that it should be `LabelType | undefined`.  But that has different semantics than an optional arg.  Remember, passing a parameter with value === undefined is different than passing no parameter.

In this particular case, both  would work, but the semantics are different.

PR #17, opened 2021-08-19
feedback: thuang-2366-feedback

tihuan:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/client/src/actions/annotation.ts
+++ b/client/src/actions/annotation.ts
@@ -1,6 +1,6 @@
-/*
-Action creators for user annotation
-*/
+/**
+ * Action creators for user annotation
+ */
 import difference from "lodash.difference";
 import pako from "pako";
 import * as globals from "../globals";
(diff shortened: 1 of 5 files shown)


client/src/actions/annotation.ts
```diff
@@ -4,21 +4,20 @@ Action creators for user annotation
...
-  async (dispatch: any, getState: any) => {
+  async (dispatch: AppDispatch, getState: GetState): Promise<void> => {
     /*
-  Add a new user-created category to the obs annotations.
-
-  Arguments:
-    newCategoryName - string name for the category.
-    categoryToDuplicate - obs category to use for initial values, or null.
-  */
+     * Add a new user-created category to the obs annotations.
```
bkmartinjr, line 16: If you reformat comments, please convert to tsdoc style:
* comment goes before the function definition, not inside
* it starts with a `/**` not `/*`
* add `@params`
etc.

bkmartinjr (approved):
LGTM

Just a nit over tsdoc comment style

tihuan:
Ooh that's right! Just updated to use the right format 🙆‍♂️ Thanks, Bruce!

tihuan:
Asked for review again just in case lol Thank you!

client/src/actions/annotation.ts
```diff
@@ -1,24 +1,22 @@
...
 import pako from "pako";
 import * as globals from "../globals";
+import { AppDispatch, GetState } from "../reducers";
 import { MatrixFBS, AnnotationsHelpers } from "../util/stateManager";
 
 const { isUserAnnotation } = AnnotationsHelpers;
 
+/**
+ * Add a new user-created category to the obs annotations.
+ * @arg newCategoryName - string name for the category.
```
bkmartinjr, line 14: should be `@param` and not `@arg`

See https://tsdoc.org/pages/tags/param/

bkmartinjr: Also see https://tsdoc.org/pages/tags/returns/

bkmartinjr (approved):
just a comment format nit, otherwise LGTM

client/src/actions/annotation.ts
tihuan: Ahh only jsdoc has `@arg` alias

tihuan: This function returns `Promise<void>`, so I suppose we don't need `@returns`?

bkmartinjr: It is certainly optional, but my _personal opinion_ is that being explicit is very helpful - so I bias towards always specifying `@returns`.

tihuan: Ahh yeah that sounds good 👍 Just added `@returns An empty promise.`!

PR #18, opened 2021-08-20
Reducer typing, pass #1

colinmegill:
Annotations reducer, genesets, pointdilation


client/src/reducers/annotations.ts
```diff
@@ -29,16 +29,27 @@ const Annotations = (
...
+  }
 ) => {
   switch (action.type) {
     case "configuration load complete": {
       const dataCollectionName =
         action.config.parameters?.["annotations-data-collection-name"] ?? null;
       const dataCollectionNameIsReadOnly =
         (action.config.parameters?.[
-          "annotations-data-collection-is-read-only"
+          "annotations-data-collection-is-read-only" /* FIXME typescript */
```
colinmegill, line 52: This line has the following error:

Element implicitly has an 'any' type because expression of type '"annotations-data-collection-is-read-only"' can't be used to index type '{ "annotations-data-collection-name": string; user_annotation_collection_name_enabled: boolean; annotations_genesets_name_is_read_only: boolean; }'.
  Property 'annotations-data-collection-is-read-only' does not exist on type '{ "annotations-data-collection-name": string; user_annotation_collection_name_enabled: boolean; annotations_genesets_name_is_read_only: boolean; }'.ts(7053)

seve: It's `-` vs` _`

seve (changes requested):
We're going to want to export the state and action types for use in other places. 

Take a look at these reducers for reference:

https://github.com/chanzuckerberg/single-cell-explorer/blob/08b03ace60f1ea411f98b82622dbf67b23a1cee3/client/src/reducers/continuousSelection.ts

https://github.com/chanzuckerberg/single-cell-explorer/blob/08b03ace60f1ea411f98b82622dbf67b23a1cee3/client/src/reducers/userInfo.ts

bkmartinjr (changes requested):
Primary issue I see was already noted by @seve - all reducers need to export types that define both their state and their action.

For example, annotations.ts should:
* define exported types for `action` and `state`
* use those in the reducer function declaration

PR #19, opened 2021-08-26
Thuang 10 type state manager

tihuan:
This PR covers `client/src/util/stateManager` with commits in diff! PTAL thank you 🙏

### DIFF
--- a/client/__tests__/util/stateManager/colorHelpers.test.ts
+++ b/client/__tests__/util/stateManager/colorHelpers.test.ts
@@ -95,8 +95,9 @@ describe("categorical color helpers", () => {
     const data = obsDataframe.col("categoricalColumn").asArray();
     const cats = schema.annotations.obsByName.categoricalColumn.categories;
     for (let i = 0; i < schema.dataframe.nObs; i += 1) {
-      // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-      expect(makeScale(ct.rgb[i])).toEqual(ct.scale(cats.indexOf(data[i])));
+      expect(makeScale(ct.rgb[i])).toEqual(
+        ct?.scale && ct?.scale(cats.indexOf(data[i]))
+      );
     }
   });
 
@@ -113,22 +114,27 @@ describe("categorical color helpers", () => {
     const data = obsDataframe.col("categoricalColumn").asArray();
     const cats = schemaClone.annotations.obsByName.categoricalColumn.categories;
     for (let i = 0; i < schemaClone.dataframe.nObs; i += 1) {
-      // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-      expect(makeScale(ct.rgb[i])).toEqual(ct.scale(cats.indexOf(data[i])));
+      expect(makeScale(ct.rgb[i])).toEqual(
+        ct?.scale && ct?.scale(cats.indexOf(data[i]))
+      );
     }
   });
 
   test("user defined color order", () => {
     const cats = schema.annotations.obsByName.categoricalColumn.categories;
+
     const shuffleCats = shuffle(
       Array.from(schema.annotations.obsByName.categoricalColumn.categories)
     );
+
     const userDefinedColorTable = {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-      categoricalColumn: shuffleCats.reduce((acc: any, label: any) => {
-        acc[label] = randRGBColor();
-        return acc;
-      }, {}),
+      categoricalColumn: shuffleCats.reduce(
+        (acc: { [label: string]: string }, label: string) => {
+          acc[label] = randRGBColor();
+          return acc;
+        },
+        {}
+      ),
     };
 
     const userColors = loadUserColorConfig(userDefinedColorTable);
@@ -139,15 +145,13 @@ describe("categorical color helpers", () => {
       "categoricalColumn",
       obsDataframe,
       schema,
-      // @ts-expect-error ts-migrate(2345) FIXME: Argument of type '{}' is not assignable to paramet... Remove this comment to see the full error message
       userColors
     );
     expect(ct).toBeDefined();
     const data = obsDataframe.col("categoricalColumn").asArray();
     for (let i = 0; i < schema.dataframe.nObs; i += 1) {
       expect(makeScale(ct.rgb[i])).toEqual(
-        // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-        ct.scale(cats.indexOf(data[i])).toString()
+        ct?.scale && ct?.scale(cats.indexOf(data[i])).toString()
       );
     }
   });
--- a/client/__tests__/util/stateManager/colorHelpers.test.ts
+++ b/client/__tests__/util/stateManager/colorHelpers.test.ts
@@ -95,9 +95,7 @@ describe("categorical color helpers", () => {
     const data = obsDataframe.col("categoricalColumn").asArray();
     const cats = schema.annotations.obsByName.categoricalColumn.categories;
     for (let i = 0; i < schema.dataframe.nObs; i += 1) {
-      expect(makeScale(ct.rgb[i])).toEqual(
-        ct?.scale && ct?.scale(cats.indexOf(data[i]))
-      );
+      expect(makeScale(ct.rgb[i])).toEqual(ct?.scale?.(cats.indexOf(data[i])));
     }
   });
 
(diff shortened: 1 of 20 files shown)


client/__tests__/util/stateManager/colorHelpers.test.ts
```diff
@@ -113,22 +114,27 @@ describe("categorical color helpers", () => {
     const data = obsDataframe.col("categoricalColumn").asArray();
     const cats = schemaClone.annotations.obsByName.categoricalColumn.categories;
     for (let i = 0; i < schemaClone.dataframe.nObs; i += 1) {
-      // @ts-expect-error ts-migrate(2722) FIXME: Cannot invoke an object which is possibly 'undefin... Remove this comment to see the full error message
-      expect(makeScale(ct.rgb[i])).toEqual(ct.scale(cats.indexOf(data[i])));
+      expect(makeScale(ct.rgb[i])).toEqual(
+        ct?.scale && ct?.scale(cats.indexOf(data[i]))
```
seve, line 118: why not:
```suggestion
        ct?.scale?.(cats.indexOf(data[i]))
```

client/src/actions/index.ts
```diff
@@ -30,12 +33,15 @@ function setGlobalConfig(config: any) {
 /*
 return promise fetching user-configured colors
 */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-async function userColorsFetchAndLoad(dispatch: any) {
+async function userColorsFetchAndLoad(
+  dispatch: AppDispatch
+): Promise<{ type: string; userColors: ConvertedUserColors }> {
   return fetchJson("colors").then((response) =>
```
seve, line 39: call `fetchJson` with a type to fulfill the type generic, this will guarantee the return type instead of casting

client/src/reducers/genesets.ts
```diff
@@ -24,9 +24,28 @@
...
+
+interface Geneset {
+  genesetName: string;
+  genesetDescription: string;
+  genes: Map<Gene["geneSymbol"], Gene>;
+}
+
+export type Genesets = Map<Geneset["genesetName"], Geneset>;
+
+interface State {
```
seve, line 40: export as less generic `GenesetsState`

seve: Also type action

client/src/util/stateManager/colorHelpers.ts
```diff
@@ -1,41 +1,47 @@
...
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  colorMode: any,
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  colorByAccessor: any,
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  schema: any,
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  genesets: any
-) {
+  colorMode: Action["type"],
```
seve, line 28: Why `Action["type"]`? Isn't this just `any`

client/src/util/stateManager/colorHelpers.ts
```diff
@@ -186,55 +199,65 @@ function _createUserColors(
...
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-function createRgbArray(data: any, colors: any) {
+function createRgbArray(data: DataframeValueArray, colors?: CategoryColors) {
   const rgb = new Array(data.length);
+
+  if (!colors) {
+    throw new Error("`colors` is undefined");
+  }
+
```
seve, line 247: What case were we seeing `createRgbArray` called with a single arg?

client/src/util/stateManager/controlsHelpers.ts
```diff
@@ -31,107 +36,80 @@ Remember that option values can be ANY js type, except undefined/null.
...
-    unused.sort((a, b) => colIndex.getOffset(a) - colIndex.getOffset(b));
-    const numToDrop =
-      unused.length < numOverWatermark ? unused.length : numOverWatermark;
-    for (let i = 0; i < numToDrop; i += 1) {
-      varData = varData.dropCol(unused[i]);
-    }
-  }
-  return varData;
+export function createCategoricalSelection(names: string[]): {
+  [key: string]: Map<unknown, unknown>;
```
seve, line 112: Based off of [this comment describing CategoricalSelection](https://github.com/chanzuckerberg/single-cell-explorer/blob/8d21a5553ef33f37224b256fcf3f094efac8d8d7/client/src/reducers/categoricalSelection.ts#L3-L13), the Map can be typed so:
```suggestion
  [key: string]: Map<string, boolean>;
```

seve (changes requested):
A few questions/comments

client/__tests__/util/stateManager/colorHelpers.test.ts
tihuan: Thanks for flagging! `scale` is a function, so we need to check the function exists before invoking it!

Unfortunately we can't do `ct?.scale?(cats.indexOf(data[i]))`, because the compiler thinks it's ternary operation or something -.-

client/src/actions/index.ts
tihuan: Ahh great idea! Thank you!

client/src/reducers/genesets.ts
tihuan: Ah I didn't export State because no one is importing it at the moment, maybe we can keep it not exported until a callsite actually needs it?

Just added `AnyAction` for `action` 🙆‍♂️

client/src/util/stateManager/colorHelpers.ts
tihuan: Technically `Action["type"]` is generic and can be strings or symbols. I opted to use `Action["type"]` here instead of `string` because I wanted to signal that `colorMode` comes from action type directly, which `string` can't convey such relationship. What do you think?

client/src/util/stateManager/colorHelpers.ts
```diff
@@ -186,55 +199,65 @@ function _createUserColors(
...
   const scale = d3
     .scaleSequential(interpolateRainbow)
-    .domain([0, categories.length]);
+    .domain([0, categories?.length || 0]);
 
   /* pre-create colors - much faster than doing it for each obs */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  const colors = categories.reduce((acc: any, cat: any, idx: any) => {
-    acc[cat] = parseRGB(scale(idx));
+  const colors = categories?.reduce((acc: CategoryColors, cat, idx) => {
```
tihuan, line 227: this `colors` can be `undefined`

client/src/util/stateManager/colorHelpers.ts
tihuan: Great question! `colors` is optional here because callsite could be undefined as `colors`. E.g., https://github.com/chanzuckerberg/single-cell-explorer/pull/19/files/bfaa6ac79d14925936517da37631bc6700bf5538#r705659559

client/src/util/stateManager/controlsHelpers.ts
tihuan: Oh thank you for linking that! Updated 🙆‍♂️

tihuan:
Just updated the PR! Thanks so much for all the helpful feedback, @seve 🙏 🙏 🙏 !

client/__tests__/util/stateManager/colorHelpers.test.ts
seve: It's ?.() not ?(), should work we use it in other places. https://github.com/chanzuckerberg/cellxgene/blob/eaae6df5e34682e5f96aadc3f1d7fc72ba342ca1/client/src/components/continuousLegend/index.js#L140

client/src/util/stateManager/colorHelpers.ts
seve: Since colorMode is passed from reducer state and not the action, shouldn't it be typed as ColorState["colorMode"] ?

client/__tests__/util/stateManager/colorHelpers.test.ts
tihuan: OH! Totally didn't know that's available 😆 Thank you! Updated 👍

client/src/util/stateManager/colorHelpers.ts
tihuan: Ooh great point! Added `ColorsState` 👌 Thank you!

seve (approved):

tihuan:
Hi @bkmartinjr ! Wanted to check if you prefer to wait for your review before merging this PR 🙏 Thanks so much!

tihuan:
@maniarathi Actually Bruce is backlogged with other higher pri tasks, are you okay with us merging this PR without his review? Thank you!

tihuan:
I'll merge first and ask Bruce to review later when he's less busy! Thank you!

Issue #35, opened 2021-09-01
Use type guards to insure type instead of casting


seve:
https://github.com/chanzuckerberg/single-cell-explorer/blob/8d1e6b49f867f3970b5b674b69e0948cf6879a0b/client/src/annoMatrix/loader.ts#L234-L234

atolopko-czi:
@seve how important is this? how much work is it? ty!

Issue #9, opened 2021-08-19
add TS typing to all reducers


bkmartinjr:

tihuan:
@atarashansky I added you as owner since you're just killing it lol Thank you!

PR #36, opened 2021-09-02
chore: type centroid util

seve:
Type the centroid util and call sites as well as needed imports.

Note:

I did some retyping of the `CategeoricalColumnSchema` to make the categories field required on columns with type `categorical'.  We'll have to do some tuning (#35) to reduce casting in favor of type guarding.

I also had to untype `nextSharedState` (it was typing as RootState) as I believe it opens up a self reference .  This should be fixed in #9  cc: @colinmegill


client/src/util/stateManager/schemaHelpers.ts
```diff
@@ -124,16 +132,15 @@ export function removeObsAnnoCategory(
 export function addObsAnnoCategory(schema: any, name: any, category: any) {
   /* add a category to a categorical annotation */
   const categories = schema.annotations.obsByName[name]?.categories;
-  if (!categories)
```
tihuan, line 127: ah are we moving this runtime check because `categories` is guaranteed somehow?

tihuan (approved):
LGTM!! Thanks for updating the types, @seve 🙏 Just one question! Thank you!

client/src/util/centroid.ts
```diff
@@ -16,22 +21,22 @@ import {
...
-  schema: any,
+  schema: Schema,
   categoryName: string,
   categoryDf: Dataframe,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  layoutChoice: any,
+  layoutChoice: LayoutChoiceState,
   layoutDf: Dataframe
-) => {
+): Map<string, CentroidCoordinateObject> => {
```
bkmartinjr, line 39: we also support non-string labels.   Shouldn't this be typed:   `Map<LabelType, CentroidCoordinateObject>`?

client/src/util/centroid.ts
```diff
@@ -105,14 +111,12 @@ const getCoordinatesByLabel = (
...
-  schema: any,
+  schema: Schema,
   categoryName: string,
   categoryDf: Dataframe,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  layoutChoice: any,
+  layoutChoice: LayoutChoiceState,
   layoutDf: Dataframe
-) => {
+): Map<string, [number, number]> => {
```
bkmartinjr, line 119: See above comment about assuming all labels are strings.

bkmartinjr (changes requested):
Overall looks good, with one caveat: the centrol helpers assume all labels are strings, which is incorrect AFAIK.  We can have datasets with non-string categoricals (numbers, bools, etc).

client/src/util/centroid.ts
seve: I wonder why I didn't get a type error for that, is it just because all of those types can be cast to string?

bkmartinjr: you have it right.  A string is a LabelType, but the converse is not true.

Typing only helps you in some cases ;)

client/__tests__/util/centroid.test.ts
```diff
@@ -31,7 +31,11 @@ describe("centroid", () => {
       schema,
       "field4",
       obsAnnotations,
```
bkmartinjr, line 33: Optional:  given the confusion around the type of an obsAnnnotation, you may want to add tests which exercise the other label types (eg, number or boolean or blended)

bkmartinjr:
@seve - you re-requested a review, but I don't see any changes to `util/centroid.ts` -- it still assumes that all obs categories contain only strings, as far as I can tell...

bkmartinjr (approved):
LGTM.

You might consider additional test coverage as noted inline.

PR #49, opened 2021-09-09
CXG spec doc updates for Corpora 2.0.0 schema (#2326)

atolopko-czi:
#### Reviewers
**Functional:** 
* @MDunitz 
* @bkmartinjr 

**Readability:** 
* @metakuni 

---

## Changes

* Rewrite parts related to Corpora 2.0.0 schema changes
* Minor formatting improvements
* Improved version history and explanation of what file formats exist "in the wild"
* Added a script to inspect CXG files metadata, for convenience.

## Reviewer Notes

* AFAICT, the basic CXG spec is unchanged, and the material changes I've made here are involved with the _expectations_ of what the Data Portal backend pipeline is going to produce in a CXG file's metadata (in the `__cxg_group_metadata__` group's `corpora` field), per this updated [code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d9c534968d592aa9ecf46791b111e1e6edc061ba/backend/corpora/dataset_processing/h5ad_data_file.py#L23).
* I'm not clear if hosted cellxgene actually needs any changes to handle these `corpora` field changes, but I'll assume that's work for another branch, if so.


dev_docs/cxg.md
```diff
@@ -1,5 +1,3 @@
-## UPDATE (9/30/2020): Starting today, the name Corpora will only be used as the internal project name, with cellxgene Data Portal being the official product name
-
```
atolopko-czi, line 2: Seemed old enough to now remove

dev_docs/cxg.md
```diff
@@ -14,9 +12,9 @@ CXG is a cellxgene-private data format, used for at-rest storage of annotated ma
 
 CXG is built upon the [TileDB](https://tiledb.com/) embedded database. Each CXG is a TileDB [group](https://docs.tiledb.com/main/api-usage/object-management), which in turn includes one or more TileDB multi-dimensional arrays.
 
-This document presumes familiarity with [TileDB terminology and concepts](https://docs.tiledb.com/main/), the [Corpora schema](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema.md) and its [H5AD encoding](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md), and the AnnData/H5AD data model.
+This document presumes familiarity with [TileDB terminology and concepts](https://docs.tiledb.com/main/), the [Corpora schema](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/corpora_schema.md).
 
-This document also leverages the current cellxgene schema, which is documented in the [REST API spec](./REST_API.md).
+This document also leverages the current cellxgene schema, which is documented in the [REST API spec](./REST_API.md). TODO: Update
```
atolopko-czi, line 17: - [ ] TODO

dev_docs/cxg.md
```diff
@@ -73,38 +71,38 @@ For example:
...
+If the `X` array is sparse, and the `X_col_shift` array exists, then all values in the `i`'th column were subtracted by `X_col_shift[i]`.
 
-### emb and embedding arrays
+### `emb` group and embedding arrays
 
-A CXG must have a group named `emb`, which will contain all embeddings.  Embeddings are encoded as TileDB arrays, of numeric type and shape (n_obs, >=2).  The arrays SHOULD be coerced to float32, and MUST be a numeric type.  The TileDB array name will be assumed to be the embedding name (conventionally, embedding names in CXG are _not_ prefixed with an `X_` as they are in AnnData).
+A CXG MUST have a group named `emb`, which will contain all embeddings.  Embeddings are encoded as TileDB arrays, of numeric type and shape `(n_obs, >=2)`.  The arrays MUST be a numeric type, and SHOULD be coerced to `float32`. The array values MUST NOT contain any infinite values and MUST contain at least one non-NaN value.  The TileDB array name will be assumed to be the embedding name (conventionally, embedding names in CXG are _not_ prefixed with an `X_` as they are in AnnData).
```
atolopko-czi, line 86: restate the numeric type requirement per the [CXG creation code](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d9c534968d592aa9ecf46791b111e1e6edc061ba/backend/corpora/dataset_processing/h5ad_data_file.py#L129).

dev_docs/cxg.md
```diff
@@ -119,46 +117,61 @@ This optional field contains a copy of the category color table, which MAY be us
...
-* the contents of `corpora_encoding_version` MUST be identical to the `cxg_version`, as this field is defined as the current object encoding version, *NOT* the source data encoding version.
-* the entire encoding will be JSON, rather than a hybrid Python/JSON encoding, but will otherwise follow the data structure defined by the AnnData Corpora encoding.
-* the `<obs_column>_colors` will be omitted in favor of `cxg_category_colors`
-
 ### Other Corpora fields
 
-All other Corpora schema fields will be encoded into a CXG using the conventions defined in the [Corpora Schema AnnData Implementation](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md). For example, fields in `AnnData.obs` will be encoded in the CXG `obs`array as defined [above](#obs-and-var).
+All other Corpora schema fields will be encoded into a CXG using the conventions defined in the [Corpora Schema](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/corpora_schema.md). For example, fields in `AnnData.obs` will be encoded in the CXG `obs`array as defined [above](#obs-and-var).
 
 ### Compatibility with CXG 0.1.0
```
atolopko-czi, line 161: In fact, since CXG is a private format, do we need to continue to support CXG 0.1.0 at all? If all CXG files are being updated to CXG 0.2.0, can we eliminate this section altogether?

dev_docs/cxg.md
```diff
@@ -119,46 +116,61 @@ This optional field contains a copy of the category color table, which MAY be us
...
-
 ### Other Corpora fields
 
-All other Corpora schema fields will be encoded into a CXG using the conventions defined in the [Corpora Schema AnnData Implementation](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md). For example, fields in `AnnData.obs` will be encoded in the CXG `obs`array as defined [above](#obs-and-var).
+All other Corpora schema fields will be encoded into a CXG using the conventions defined in the [Corpora Schema 2.0.0](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/corpora_schema.md). For example, fields in `AnnData.obs` will be encoded in the CXG `obs`array as defined [above](#obs-and-var).
 
 ### Compatibility with CXG 0.1.0
 
-For backwards compatibility and continuity with CXG version 0.1.0, the following MUST be implemented.
+For CXG v0.1.0 files that were created from [Corpora Schema 1.1.0](https://github.com/chanzuckerberg/single-cell-curation/blob/2dc6a6f7ea342715d8a400a3cead29c691745df7/schema/1.1.0/corpora_schema.md) H5AD files, the following WILL apply:
```
atolopko-czi, line 162: Is this line's rewrite accurate?  I was confused by the usage of "MUST".  Presumably the hosted cellxgene FE code doesn't care where the required field values are _taken from_ in an H5AD file, only that some value be provided.

dev_docs/cxg.md
atolopko-czi: Per Slack discussion cxg 0.1.0 files will persist for a while, so keeping this in here for now.

dev_docs/cxg.md
```diff
@@ -119,46 +116,61 @@ This optional field contains a copy of the category color table, which MAY be us
...
+* `X_approximate_distribution`
 
+These Corpora metadata properties MAY exist in the CXG `__cxg_group_metadata__` field named `corpora`:
+* `contributors`
+* `default_embedding`
+* `project_name`
+* `project_description`
+* `project_links`
+* `preprint_doi`
+* `publication_doi`
```
atolopko-czi, line 141: none of these are required by Schema 2.0.0, but I'm assuming they may exist by convention alone. As these are the only props that will be passed through into the CXG, I suppose it's worth listing them here.

brianraymor: This is schema 1.0.0 which embedded collection (project) information in the dataset prior to the UX. Deprecated in schema 1.1.0.  See https://github.com/chanzuckerberg/single-cell/issues/75.

The remaining 1.0.0 datasets in the corpus (can identify them by the missing cell count in _All Collections_ view will be updated during re-curation by Stanford.

atolopko-czi: Thanks, so I will remove this section. 

Separately, we may want to remove from the H5AD->CXG conversion logic [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d9c534968d592aa9ecf46791b111e1e6edc061ba/backend/corpora/common/utils/corpora_constants.py#L93). I'm assume that will be handled by later schema 1.x deprecation work..

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
+### `X` array
 
-TileDB array, with a single anonymous attribute, shape (n_obs, n_var), containing the count matrix (equivalent to the AnnData `X` array).  MUST have numeric type, and SHOULD be float32.  The TileDB schema defines type and sparsity, and both dense and sparse encoding are supported.
+TileDB array, with a single anonymous attribute, shape `(n_obs, n_var)`, containing the count matrix (equivalent to the AnnData `X` array).  MUST have numeric type, and SHOULD be `float32`.  The TileDB schema defines type and sparsity, and both dense and sparse encoding are supported.
 
-### X_col_shift
+### `X_col_shift` array
 
-Optional TileDB array, used to encode-per column offsets for column-shift sparse encoding.  The TileDB array will have a single anonymous attribute, of the same type as the X array, and shape (n_var,).  
+Optional TileDB array, used to encode per-column offsets for column-shift sparse encoding.  The TileDB array will have a single anonymous attribute, of the same type as the X array, and shape `(n_var,)`.
```
bkmartinjr, line 79: @atolopko-czi - quick note on this -- in reading the implementation (in matrix_utils.py), I note that the code currently does _not_ correctly implement the `same type as the X array` requirement.  I think this is a code bug, and that the spec is correct.

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
+TileDB array, with a single anonymous attribute, shape `(n_obs, n_var)`, containing the count matrix (equivalent to the AnnData `X` array).  MUST have numeric type, and SHOULD be `float32`.  The TileDB schema defines type and sparsity, and both dense and sparse encoding are supported.
 
-### X_col_shift
+### `X_col_shift` array
 
-Optional TileDB array, used to encode-per column offsets for column-shift sparse encoding.  The TileDB array will have a single anonymous attribute, of the same type as the X array, and shape (n_var,).  
+Optional TileDB array, used to encode per-column offsets for column-shift sparse encoding.  The TileDB array will have a single anonymous attribute, of the same type as the X array, and shape `(n_var,)`.  
 
-If the X array is sparse, and X_col_shift exists, then all values in the i'th column were subtracted by X_col_shift[i].
+If the `X` array is sparse, and the `X_col_shift` array exists, then all values in the `i`'th column were subtracted by `X_col_shift[i]`.
```
bkmartinjr, line 81: why is this unique to sparse X matrix?  Seems like it is applicable to sparse or dense X matrix.

dev_docs/cxg.md
bkmartinjr: The rewrite contains redundant "numeric type" requirements (one in the second sentence, another in the third).

I'm curious about your edits regarding non-scalars.  What is the use case for NaN's in an embedding?  Why not just say all embedding coordinates must have a finite value?

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
-### emb and embedding arrays
+### `emb` group and embedding arrays
 
-A CXG must have a group named `emb`, which will contain all embeddings.  Embeddings are encoded as TileDB arrays, of numeric type and shape (n_obs, >=2).  The arrays SHOULD be coerced to float32, and MUST be a numeric type.  The TileDB array name will be assumed to be the embedding name (conventionally, embedding names in CXG are _not_ prefixed with an `X_` as they are in AnnData).
+A CXG MUST have a group named `emb`, which will contain all embeddings.  Embeddings are encoded as TileDB arrays, of numeric type and shape `(n_obs, >=2)`.  The arrays MUST be a numeric type, and SHOULD be coerced to `float32`. The array values MUST NOT contain any infinite values and MUST contain at least one non-NaN value.  The TileDB array name will be assumed to be the embedding name (conventionally, embedding names in CXG are _not_ prefixed with an `X_` as they are in AnnData).
 
 CXG supports zero or more embeddings. Note that cellxgene currently _requires_ at least one embedding.
 
-### cxg_group_metadata
+### `cxg_group_metadata` group
```
bkmartinjr, line 89: this is not a group, but an array (as it states in the first sentence).

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
 
 Required, but empty TileDB array, used to store CXG-wide metadata.  The following fields are defined:
```
bkmartinjr, line 91: Suggest:
> ...used to store CXG-wide metadata within [array metadata](https://docs.tiledb.com/main/solutions/tiledb-embedded/api-usage/array-metadata)

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
+* `__cxg_properties__`: (optional) a dictionary containing dataset wide properties, defined below.
+* `__cxg_category_colors__`: (optional) a categorical color table, defined below.
 
-#### cxg_properties
+#### `cxg_properties` field
 
-The properties metadata dictionary contains dataset-wide properties, encoded as a JSON dictionary. Currently, the following fields are defined:
-* title: string, dataset human name (eg, "Lung Tissue")
-* about: string, fully-qualified http/https URL, linking to more information on the dataset.
+The `__cxg_properties__` metadata dictionary contains dataset-wide properties, encoded as a JSON dictionary. Currently, the following fields are defined:
```
bkmartinjr, line 98: properties of this metadata field should be noted as optional/mandatory.  Currently, I  believe they are all optional (eg, you can have a cxg_properties dict that only contains 'title')

dev_docs/cxg.md
```diff
@@ -119,46 +116,51 @@ This optional field contains a copy of the category color table, which MAY be us
...
 {
     "corpora": {
-        "version": {
-            "corpora_schema_version": "1.0.0",
-            "corpora_encoding_version": "0.1.0",
-        }
+        "schema_version": "2.0.0",
+        "title": "Analysis of 8 datasets of healthy and diseased mouse heart",
+        "X_normalization": "CPM",
+        "X_approximate_distribution": "normal
```
bkmartinjr, line 141: missing terminating quote

dev_docs/cxg.md
bkmartinjr: I don't think this is correct -- IIRC, the versions that were created with Corpora schema are CXG 0.2.0 (see version history below)

dev_docs/cxg.md
```diff
@@ -1,5 +1,3 @@
-## UPDATE (9/30/2020): Starting today, the name Corpora will only be used as the internal project name, with cellxgene Data Portal being the official product name
-
 # CXG Data Format Specification
 
 Document Status: _draft_
```
bkmartinjr, line 3: given the handful of MUST/SHOULD changes, recommend we bump the CXG version number (at least to 0.2.1)

bkmartinjr (changes requested):
handful of small comments/suggestions.  Feel free to hit me up for a re-review as warranted.

bkmartinjr:
Additional comment: it would be very helpful if the spec included a common method or process by which a CXG reader could determine:
* is the CXG derived from a Corpora schema compliant dataset?
* and if it is, which version of the Corpora schema

dev_docs/cxg.md
```diff
@@ -43,25 +40,25 @@ CXG consumers (readers) MUST be prepared to handle any legal TileDB compression,
...
 
 The CXG is a TileDB group containing all data and metadata for a single annotated matrix.  The following objects MUST be present in a CXG, except where noted as optional:
-* __obs__: a TileDB array, of shape (n_obs,), containing obs annotations, each annotation stored in a separate TileDB array attribute.
-* __var__: a TileDB array, of shape (n_var,), containing var annotations, each annotation stored in a separate TileDB array attribute.
-* __X__: a TileDB array, of shape (n_obs, n_var), with a single TileDB attribute of numeric type.
-* __X_col_shift__: (optional) TilebDB Array used in column shift encoding, shape (n_var,), dtype = X.dtype. Single unnamed numeric attribute.  
-* __emb__: a TileDB group, which in turn contains all (zero or more) embeddings.
-* __emb__/\<embedding_name\>__: a TileDB array, with a single anonymous attribute, of numeric type, and shape (n_obs, N>=2).
-* __cxg_group_metadata__: an empty TileDB array, used to store CXG-wide metadata
+* `__obs__`: a TileDB array, of shape (n_obs,), containing obs annotations, each annotation stored in a separate TileDB array attribute.
```
bkmartinjr, line 43: This quoting changes the schema definition entirely -- `obs` becomes `__obs__`.  I assume this is unintentional, and you want to drop the bolding (__) from each, or revert the back-tick-quoting.

dev_docs/cxg.md
```diff
@@ -73,38 +70,38 @@ For example:
...
 CXG supports zero or more embeddings. Note that cellxgene currently _requires_ at least one embedding.
 
-### cxg_group_metadata
+### `cxg_group_metadata` group
 
 Required, but empty TileDB array, used to store CXG-wide metadata.  The following fields are defined:
-* __cxg_version__: (required) a semver string identifying the specification version used to encode the CXG.
-* __cxg_properties__: (optional) a dictionary containing dataset wide properties, defined below.
-* __cxg_category_colors__: (optional) a categorical color table, defined below.
+* `__cxg_version__`: (required) a [semantic version](https://semver.org/) identifying the specification version used to encode the CXG.
```
bkmartinjr, line 92: these are also malformed due to the addition of the quotes.  The actual key does not include __ prefix or suffix.

Eg, it is `cxg_version`, not `__cxg_version__`

dev_docs/cxg.md
atolopko-czi: I interpret the [h5ad -> cxg implementation](https://github.com/chanzuckerberg/single-cell-data-portal/blob/d9c534968d592aa9ecf46791b111e1e6edc061ba/backend/corpora/dataset_processing/h5ad_data_file.py#L99-L111) as follows: "If the H5AD-provided `X` matrix is not sufficiently sparse, per the desired `sparse_threshold`, attempt to compute a `col_shift` array that _would_ allow it to be stored as a sparse matrix that satisfies the desired `sparse_threshold`." Does this agree with your interpretation?

If so, perhaps the spec should read:
> If the `X_col_shift` array exists, then the matrix SHOULD be a TileDB sparse matrix, and all the values in the `i`'th column MUST be add by `X_col_shift[i]` to restore the originally-provided values from the source H5AD file.

I assume one could specify a `X_col_shift` in the CXG file even when the X matrix is stored densely, though that would be silly. Thus the `SHOULD`, above.

dev_docs/cxg.md
atolopko-czi: The edits were only based upon what my reading of what the h5ad->cxg implementation is actually doing, rather than perhaps what it MUST be doing to make explorer happy.  I assumed the conversion code knew what it was doing and why 😄. So, yes, if explorer allows all float values, it doesn't seem worth documenting this as a spec constraint. Though it seems cleaner to insist on finite floats for the intended visualization purpose of these embeddings.

dev_docs/cxg.md
atolopko-czi: Thanks for the catch, fixed!  Was likely fooled by the [method name](https://github.com/chanzuckerberg/single-cell-data-portal/blob/135214e9f423d815645fb3e953dc3a222b7fd252/backend/corpora/common/utils/cxg_generation_utils.py#L12)

dev_docs/cxg.md
atolopko-czi: added

dev_docs/cxg.md
atolopko-czi: now noted all are optional (verified in explorer code this is safe)

dev_docs/cxg.md
atolopko-czi: purely to ensure reviewers are doing their job. 😉  fixed!

dev_docs/cxg.md
atolopko-czi: I'm a bit confused as to what do with this section--do we need it at all? Seems out of place for a 0.2.0 spec.  I would assume that if we have CXG 0.1.0 files currently in use, the content in this section would be better off in a separate 0.1.0 spec file, and explorer  code would simply be considered compatible with both 0.1.0 and 0.2.0.   If 0.1.0 files do not exist (contrary to `Created in early 2020, and in active use`, below), we could presumably just delete this section.

dev_docs/cxg.md
bkmartinjr: I am not sure this spec should mandate a sparsity encoding threshold, or even get into the (policy) decision of when the matrix is sparse.  So, from the viewpoint of this spec, I think it would be more accurate to say a small variation on your original theme:

> If the `X` array is sparse, and the `X_col_shift` array exists, then all values in the `i`'th column of `X` are decremented by the value of the corresponding `X_col_shift[i]`.   `X_col_shift` MAY exist if the `X` array is sparse, and when it doe not exist, has identical semantics to a column shift value of zero (0) for each column.   `X_col_shift` MUST NOT exist if the `X` array is dense.

TL;DR:
* dense must not have a col shift array
* sparse may have a col-shift array, and if it doesn't, it is identical to the semantics of a col-shift of zero

dev_docs/cxg.md
bkmartinjr: I think the real question is if there is any viable use case for an embedding algo to generate a non-finite.  i can't think of one, although I recommend validating with Amborse/Pablo/Jason.  Assuming I am correct, then I'd tighten this to "only finite float values"

dev_docs/cxg.md
bkmartinjr: I would also add a note to the effect that it must be a semver 2.0 complaint version string.

we were bit by this in an early version of CXG, where I mistakenly hard-wired a malformed version!

dev_docs/cxg.md
bkmartinjr: I think there are two useful things you could do, which would improve our code:
1. define a canonical algo to sniff the schema version out of _any_ CXG (ie, is it 2.0.0, 0.1.0, ????)
2. define what, if any, defaults should be used for critical metadata when reading old schema.  This may be a null set.
I other words, tighten up how various apps interpret CXG backwards compat.

dev_docs/cxg.md
atolopko-czi: Happy to bump it, but in fact, I was only attempting to update the doc to match what appeared to be the "intended" 0.2.0 specification, rather than trying to change the specification.  I was considering the h5ad->cxg _code_ to be the ground truth for what constituted the spec.  And as this specification is for our private use only (right?), making changes to the text for an _existing_ version of the spec seemed okay. But I can agree that any changes to the doc should reasonably result in a version bump.

dev_docs/cxg.md
atolopko-czi: indeed, I misinterpreted `__` as part of the literal names. fixed!

dev_docs/cxg.md
atolopko-czi: added bug: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-data-portal/1410

dev_docs/cxg.md
atolopko-czi: agreed, the sparsity threshold is not a spec concern, which should only care about ensuring the `X_col_shift` is used correctly.

will incorporate your rewrite, which additionally stipulates "dense implies no x_col_shift"

dev_docs/cxg.md
atolopko-czi: taking this to [slack](https://czi-sci.slack.com/archives/C024HCSH9PT/p1631579063117000)

dev_docs/cxg.md
atolopko-czi: specified semver 2.0 explicitly.

dev_docs/cxg.md
```diff
@@ -119,46 +120,51 @@ This optional field contains a copy of the category color table, which MAY be us
...
+
+The [Corpora Schema 2.0.0](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/corpora_schema) defines a set of metadata and encoding conventions for annotated matrices.  When a Corpora dataset is encoded as a CXG, the following WILL apply.   
 
 ### Corpora metadata property
 
-A CXG containing a Corpora dataset will contain a property in the __cxg_group_metadata__ field named `corpora`. The value will be a JSON encoded string, which in turn contains all properties defined in the [Corpora AnnData uns](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md#uns) container. For example:
+A CXG containing a Corpora dataset WILL contain a property in the `cxg_group_metadata` field named `corpora`. The value WILL be a JSON encoded string, which in turn contains all properties defined in the [Corpora Schema 2.0.0 "uns"](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/schema.md#uns-dataset-metadata) AnnData container. The entire encoding WILL be JSON, rather than a hybrid Python/JSON encoding, but will otherwise follow the data structure defined by the Corpora schema.
```
metakuni, line 129: Super nit: `` ` `` instead of `"` around `uns`?

dev_docs/cxg.md
```diff
@@ -119,46 +120,51 @@ This optional field contains a copy of the category color table, which MAY be us
...
+* The [Corpora Schema 1.1.0 SUMMARY `project_link`](https://github.com/chanzuckerberg/single-cell-curation/blob/2dc6a6f7ea342715d8a400a3cead29c691745df7/schema/1.1.0/corpora_schema.md#presentation-hints), MAY be saved in the `cxg_properties.about` field.
 
 Where these values differ in the final CXG, the `cxg_properties` values WILL take precedence.
 
 ## CXG Version History
 
-There were several ad hoc version of CXG created prior to this spec.  This describes the _proposed_ next version of CXG, which incoporates support for Corpora schema semantics.  Prior verisons:
-* _unnamed_ - an unnamed development version. Did not include explicit versioning support in the data model, but can be detected by the absence of __cxg_group_metadata__ and any version property. Created in early 2020, and not actively used in production
+There were several ad hoc version of CXG created prior to this spec.  This describes the _proposed_ next version of CXG, which incoporates support for Corpora schema semantics.  Prior versions:
```
metakuni, line 167:
```suggestion
There were several ad hoc versions of CXG created prior to this spec.  This describes the _proposed_ next version of CXG, which incorporates support for Corpora schema semantics.  Prior versions:
```

dev_docs/cxg.md
```diff
@@ -32,9 +29,9 @@ The `cxg` prefix is used for CXG-specific names.
 
 ### Encoding Data With TileDB Arrays
 
-The TileDB array schema authoritatively defines the characteristics of each array (eg, the type of `X` is defined by [`X.schema`](https://tiledb-inc-tiledb-py.readthedocs-hosted.com/en/stable/python-api.html#tiledb.libtiledb.Array.schema)). In some cases, additional metadata is required for the CXG, and is attched to the array using the TileDB [array metadata](https://docs.tiledb.com/main/basic-concepts/array-metadata) capability.
+The TileDB array schema authoritatively defines the characteristics of each array (eg, the type of `X` is defined by [`X.schema`](https://tiledb-inc-tiledb-py.readthedocs-hosted.com/en/stable/python-api.html#tiledb.libtiledb.Array.schema)). In some cases, additional metadata is required for the CXG, and is attached to the array using the TileDB [array metadata](https://docs.tiledb.com/main/basic-concepts/array-metadata) capability.
```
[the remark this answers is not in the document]
metakuni:
```suggestion
The TileDB array schema authoritatively defines the characteristics of each array (eg, the type of `X` is defined by [`X.schema`](https://tiledb-inc-tiledb-py.readthedocs-hosted.com/en/stable/python-api.html#tiledb.libtiledb.Array.schema)). In some cases, additional metadata is required for the CXG, and is attached to the array using the TileDB [array metadata](https://docs.tiledb.com/main/basic-concepts/data-format#array-metadata) capability.
```

dev_docs/cxg.md
```diff
@@ -119,46 +120,51 @@ This optional field contains a copy of the category color table, which MAY be us
 
 ## Corpora Schema Encoding
 
-The [Corpora schema](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema.md) and [Corpora AnnData encoding](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md) define a set of metadata and encoding conventions for annotated matrices.  When a Corpora dataset is encoded as a CXG, the following shall apply.
+Note: This section is technically not part of the CXG specification, but rather provides an expectation of what metadata will be available in a CXG file that is converted from Corpora Schema-compliant H5AD file; we use "WILL" rather than "MUST" to document these expectations.
+
+The [Corpora Schema 2.0.0](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/corpora_schema) defines a set of metadata and encoding conventions for annotated matrices.  When a Corpora dataset is encoded as a CXG, the following WILL apply.
```
metakuni, line 125:
```suggestion
The [Corpora Schema 2.0.0](https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/schema.md) defines a set of metadata and encoding conventions for annotated matrices.  When a Corpora dataset is encoded as a CXG, the following WILL apply.   
```

metakuni (approved):

dev_docs/cxg.md
```diff
@@ -43,25 +40,25 @@ CXG consumers (readers) MUST be prepared to handle any legal TileDB compression,
...
+* `X_col_shift`: (optional) TileDB array used in column shift encoding, shape (n_var,), dtype = X.dtype. Single unnamed numeric attribute.  
+* `emb`: a TileDB group, which in turn contains all (zero or more) embeddings.
+* `emb/<EMBEDDING_NAME>`: a TileDB array, with a single anonymous attribute, of numeric type, and shape (n_obs, N>=2).
+* `cxg_group_metadata`: an empty TileDB array, used to store CXG-wide metadata
 
-### obs and var
+### `obs` and `var` arrays
 
-All per-observation (obs) and per-feature (var) data is encoded in a TileDB array named `obs` and `var` respectively, with shape (n_obs,) and (n_var,).  Each TileDB array has an array attribute for each obs/var column.  All TileDB array attributes will have the same type and value as the original data, eg, float32, with the following exceptions:
+All per-observation (obs) and per-feature (var) data is encoded in a TileDB array named `obs` and `var` respectively, with shape `(n_obs,)` and `(n_var,)`.  Each TileDB array has an array attribute for each obs/var column.  All TileDB array attributes will have the same type and value as the original data, eg, float32, with the following exceptions:
```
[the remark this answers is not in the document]
atolopko-czi: yep, done!

dev_docs/cxg.md
atolopko-czi: updated

dev_docs/cxg.md
atolopko-czi: updaated

dev_docs/cxg.md
atolopko-czi: Updated date, but I'm not going to update the version: See below [comment](https://github.com/chanzuckerberg/single-cell-explorer/pull/49/files#r707791218) and further explanation that I've since added in `CXG Version History` section.

dev_docs/cxg.md
atolopko-czi: Not going to bump the version, since this 0.2.0 spec defines the current state of affairs, in terms of real data in the wild. There are files that now have 0.2.0 cxg version and conform to this spec as stated. We'll need to be more careful in the future. But honestly, all of the substantive changes are in the `corpora` metadata, which is not even part of the CXG spec technically.

dev_docs/cxg.md
atolopko-czi: TODO: I will relax this to state what is actually allowed by the explorer code, and if we tighten the constraints, we'll want to bump  CXG version along with the Explorer code changes.

dev_docs/cxg.md
atolopko-czi: Leaving this section as now edited, since it's still relevant for the remaining `/e/` datasets that are in CXG 0.1 format. I added a note above regarding this. We can remove this section once all the CXG 0.1 files are wiped from existence (at least in the production env). I don't think we should worry heavily about backwards compatibility at this point.

Added more info about CXG & Corpora schema detection in the Versions section, below.

dev_docs/cxg.md
bkmartinjr: works for me

bkmartinjr (approved):
LGTM!

PR #52, opened 2021-09-09
Dunitz/tombstone redirect

MDunitz:
#### Reviewers
**Functional:** 
@ebezzi  @MDunitz 

**Readability:** 
@dimple-jethani 

---

## Changes
- add `TombstoneException` handler to redirect user to the collections page with query param `tombstoned_dataset_id=<dataset_id>` when a user attempts to view a tombstoned dataset.

(Note: The video and actual behavior differs because of the cache delay of 5 minutes. After deleting a datasets, it will take ~5 minutes for the redirect to work since we cache dataset metadata with a TTL of 5 min.)
https://user-images.githubusercontent.com/75650148/133624675-b3e3bac9-851d-4d85-a19b-5d39e4c5074a.mov


server/app/app.py
```diff
@@ -67,6 +67,11 @@ def cache_control_always(**cache_kwargs):
     return _cache_control(True, **cache_kwargs)
 
 
+@webbp.errorhandler(RequestException)
+def handle_request_exception(error):
```
leslieypark, line 71: Removing
```
@webbp.errorhandler(TombstoneException)
def handle_tombstone_exception(error):
    return redirect(f"{current_app.app_config.server_config.get_web_base_url()}/collections/{error.collection_id}?tombstoned_dataset_id={error.dataset_id}") # noqa E501
```

Still works. Doesn't seem necessary. `CellxgeneException` also is not "registered", but the errors defined with this exception class are still functional.

server/app/app.py
```diff
@@ -94,11 +99,9 @@ def dataset_index(url_dataroot=None, dataset=None):
         return common_rest.abort_and_log(
             e.status_code, f"Invalid dataset {dataset}: {e.message}", loglevel=logging.INFO, include_exc_info=True
         )
-
-
-@webbp.errorhandler(RequestException)
-def handle_request_exception(error):
```
MDunitz, line 100: I think we may want to keep this to ensure that failed requests are logged?

leslieypark: This line (for `handle_request_exception` was just moved up - see starting line 70

server/data_common/dataset_metadata.py
```diff
@@ -71,7 +71,16 @@ def get_dataset_metadata_for_explorer_location(dataset_explorer_location: str, a
         dataset_metadata = request_dataset_metadata_from_data_portal(
             data_portal_api_base=app_config.server_config.data_locator__api_base, explorer_url=explorer_url_path
         )
+
         if dataset_metadata:
+            if dataset_metadata['tombstoned']:
+                dataset_id = dataset_metadata['dataset_id']
+                collection_id = dataset_metadata['collection_id']
+                msg = f"Dataset {dataset_id} from collection {collection_id} has been tombstoned and is no " \
```
ebezzi, line 79: I believe for strings the default way to do this is with triple quotes.

server/data_common/dataset_metadata.py
```diff
@@ -71,7 +71,16 @@ def get_dataset_metadata_for_explorer_location(dataset_explorer_location: str, a
         dataset_metadata = request_dataset_metadata_from_data_portal(
             data_portal_api_base=app_config.server_config.data_locator__api_base, explorer_url=explorer_url_path
         )
+
         if dataset_metadata:
+            if dataset_metadata['tombstoned']:
```
ebezzi, line 76: Not sure why the linter doesn't complain, but I think we're enforcing double quotes here.

server/tests/unit/common/test_api.py
```diff
@@ -510,6 +510,7 @@ def setUpClass(cls, mock_get):
         cls.config = AppConfig()
         cls.config.update_server_config(
             data_locator__api_base=cls.data_locator_api_base,
+            app__web_base_url="https://cellxgene.staging.single-cell.czi.technology.com",
```
ebezzi, line 513: Are we sure we want to hardcode this here (even if it's a test)? I think this should be available somewhere else. If not, I would at least try to parametrize the environment.

ebezzi (approved):
Approved with a few nitpicks.

server/tests/unit/common/test_api.py
leslieypark: No different from line 509 above?

server/data_common/dataset_metadata.py
leslieypark: I think it's cleaner here without using triple quotes, since that requires dedent-ting and removal of return line

PR #71, opened 2021-09-17
feat: include localhost and canary env as allowed origins for staging backend

MDunitz:
#### Reviewers
**Functional:** 
- @tihuan 
**Readability:** 
- @blrnw3 
---

## Changes
- add localhost and canary env as allowed origins

### DIFF
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -10,7 +10,6 @@ import logging
 from flask_talisman import Talisman
 from flask_cors import CORS
 
-
 if os.path.isdir("/opt/python/log"):
     # This is the standard location where Amazon EC2 instances store the application logs.
     logging.basicConfig(
@@ -80,8 +79,13 @@ class WSGIServer(Server):
         web_base_url = server_config.get_web_base_url()
         if web_base_url:
             web_base_url_parse = urlparse(web_base_url)
-            allowed_origin = f"{web_base_url_parse.scheme}://{web_base_url_parse.netloc}"
-            CORS(app, supports_credentials=True, origins=allowed_origin)
+            allowed_origins = [f"{web_base_url_parse.scheme}://{web_base_url_parse.netloc}"]
+            if os.getenv('DEPLOYMENT_STAGE') in ["Staging"]:
+                allowed_origins.extend([
+                    "https://canary-cellxgene.dev.single-cell.czi.technology/",
+                    r"^http://localhost:\d+",
+                ])
+            CORS(app, supports_credentials=True, origins=allowed_origins)
 
         Talisman(
             app,
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -80,7 +80,7 @@ class WSGIServer(Server):
         if web_base_url:
             web_base_url_parse = urlparse(web_base_url)
             allowed_origins = [f"{web_base_url_parse.scheme}://{web_base_url_parse.netloc}"]
-            if os.getenv('DEPLOYMENT_STAGE') in ["Staging"]:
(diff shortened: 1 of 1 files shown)


server/eb/app.py
```diff
@@ -80,8 +79,13 @@ def _before_adding_routes(app, app_config):
         web_base_url = server_config.get_web_base_url()
         if web_base_url:
             web_base_url_parse = urlparse(web_base_url)
-            allowed_origin = f"{web_base_url_parse.scheme}://{web_base_url_parse.netloc}"
-            CORS(app, supports_credentials=True, origins=allowed_origin)
+            allowed_origins = [f"{web_base_url_parse.scheme}://{web_base_url_parse.netloc}"]
+            if os.getenv('DEPLOYMENT_STAGE') in ["Staging"]:
```
tihuan, line 83: Ah I found [here](https://github.com/chanzuckerberg/single-cell-infra/blob/17047ab5bcfdff5a6e1c3e62eeb791628cf83c01/terraform/envs/staging/dp-staging/variables.tf#L2) uses `staging` instead of `Staging`, but not sure if that's where we're getting `DEPLOYMENT_STAGE` from 😆

server/eb/app.py
```diff
@@ -80,8 +79,13 @@ def _before_adding_routes(app, app_config):
...
+                allowed_origins.extend([
+                    "https://canary-cellxgene.dev.single-cell.czi.technology/",
```
tihuan, line 85: The auto-deployed canary env is https://cellxgene.canary.single-cell.czi.technology. Should we use that instead? Thank you!

server/eb/app.py
MDunitz: I tested by deploying to staging and "Staging" seemed to work but I can include both for safety

server/eb/app.py
MDunitz: Sorry where are you seeing that? `https://canary-cellxgene.dev.single-cell.czi.technology/` is set as the web base url in the config so I think thats what we should be using?

server/eb/app.py
MDunitz:
```suggestion
            if os.getenv('DEPLOYMENT_STAGE') in ["Staging", "staging"]:
```

server/eb/app.py
tihuan: This is the auto deploy GH flow I was referring to!

https://github.com/chanzuckerberg/single-cell-infra/runs/3547390406?check_suite_focus=true

Screen Shot 2021-09-17 at 2 20 57 PM

I also see `canary` and `canary-hook`, so not sure if we only need one or both 😆 

https://github.com/chanzuckerberg/single-cell-infra/actions?page=3

Screen Shot 2021-09-17 at 2 21 48 PM

MDunitz: So canary hook is sent from the explorer repo to the single-cell-infra repo when code is merged into the main-canary branch. Single-cell-infra then kicks off the deployment in the canary step

tihuan: Ahh got it, that's why both are needed! Thanks, Madison 🙆‍♂️ 

So you're saying https://canary-cellxgene.dev.single-cell.czi.technology/ is the right URL, and the smoke test is just pointing to the wrong one (https://cellxgene.canary.single-cell.czi.technology?

tihuan (approved):
LGTM! I think the discussion for Canary env is not a blocker, since we still need to hear back from Emanuele + Infra on rdev anyway 😆 Thank you!! 🚀 🚀

server/eb/app.py
MDunitz: yea

PR #74, opened 2021-09-21
fix: 73 - add 2.0.0 to the recognized FE versions

ebezzi:
#### Reviewers
**Functional:** 
@tihuan 

Issue [#73 ](https://github.com/chanzuckerberg/single-cell-explorer/issues/73)

## Changes
- Required, otherwise newly uploaded datasets will show a weird behavior (namely, the title is wrong).

### DIFF
--- a/client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
+++ b/client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
@@ -7,6 +7,7 @@ import Logo from "../framework/logo";
 import Truncate from "../util/truncate";
 import InfoDrawer from "../infoDrawer/infoDrawer";
 import InformationMenu from "./infoMenu";
+import { checkValidVersion } from "../util/version";
 
 const DATASET_TITLE_FONT_SIZE = 14;
 
@@ -14,10 +15,7 @@ const DATASET_TITLE_FONT_SIZE = 14;
 @connect((state) => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
   const { corpora_props: corporaProps } = (state as any).config;
-  const correctVersion =
-    ["1.0.0", "1.1.0", "2.0.0"].indexOf(
-      corporaProps?.version?.corpora_schema_version
-    ) > -1;
+  const correctVersion = checkValidVersion(corporaProps);
   return {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
     datasetTitle: (state as any).config?.displayNames?.dataset ?? "",
--- a/client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
+++ b/client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
@@ -15,8 +15,9 @@ const DATASET_TITLE_FONT_SIZE = 14;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
   const { corpora_props: corporaProps } = (state as any).config;
   const correctVersion =
-    ["1.0.0", "1.1.0"].indexOf(corporaProps?.version?.corpora_schema_version) >
-    -1;
+    ["1.0.0", "1.1.0", "2.0.0"].indexOf(
+      corporaProps?.version?.corpora_schema_version
+    ) > -1;
   return {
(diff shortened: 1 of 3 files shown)


client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
```diff
@@ -7,16 +7,15 @@ import Logo from "../framework/logo";
...
 const DATASET_TITLE_FONT_SIZE = 14;
 
 // @ts-expect-error ts-migrate(1238) FIXME: Unable to resolve signature of class decorator whe... Remove this comment to see the full error message
 @connect((state) => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
   const { corpora_props: corporaProps } = (state as any).config;
-  const correctVersion =
-    ["1.0.0", "1.1.0"].indexOf(corporaProps?.version?.corpora_schema_version) >
-    -1;
+  const correctVersion = checkValidVersion(corporaProps);
```
tihuan, line 18: prefer prefix with `is` 🙏 `const isCorrectVersion = ...`

client/src/components/util/version.ts
```diff
@@ -0,0 +1,11 @@
+export function checkValidVersion(corporaProps: any) {
+  const isValidVersionOne =
+    ["1.0.0", "1.1.0"].indexOf(corporaProps?.version?.corpora_schema_version) >
```
tihuan, line 3: super nit:

Hoist constants out of the function declaration, so we don't need to create new arrays every time:

```ts
const VERSION_ONES = ["1.0.0", "1.1.0"];
const VERSION_TWOS = ["2.0.0"];

export function checkValidVersion(corporaProps: any) {
  if (!corporaProps) return false;

  const { version, schema_version } =  corporaProps;

  const isValidVersionOne = VERSION_ONES.includes(version?.corpora_schema_version);
  const isValidVersionTwo = VERSION_TWOS.includes(schema_version);

  return isValidVersionOne || isValidVersionTwo; 
}
```

tihuan (approved):
Love the extraction! Thanks so much for fixing this, @ebezzi 🙏 Just two tiny nits!

client/src/components/leftSidebar/topLeftLogoAndTitle.tsx
tihuan: or maybe `const isValidVersion = checkValidVersion(corporaProps);` ?

client/src/components/util/version.ts
ebezzi: Good point, thanks!

atolopko-czi (changes requested):
When the CXG is created, why isn't the Schema 2.0.0-provided title being put into the CXG at `cxg_group_metadata | cxg_properties | title`?  It would seem cleaner to simply read it from a single CXG location unconditionally. More generally, it would be cleaner if Explorer never needed to care about the Schema version and instead could just rely upon the CXG file version.  So I'm questioning whether this fix should actually be a Data Portal CXG transformer fix.  Are there other places in the Explorer that we the Schema version is depended upon?

ebezzi:
@atolopko-czi The title is there. I believe this check was to prevent legacy datasets (the `/d/` nodes) from breaking, since they didn't have a title. There is no difference in behavior between 1.1 and 2.0 with respect to this change.

tihuan (approved):
Amazinggg! LGTM 🚀 🚀 🚀 Thanks so much again, @ebezzi 🌟

atolopko-czi:
Discussed w/ @ebezzi  that we should in fact be able to eliminate h5ad schema version checks in this code, but we'll merge this as is for. I'll file a followup story to replace the version check with a robust read of `cxg_group_metadata | cxg_properties | title`, since the CXGs for Schema-compliant H5AD files also put the title there.

seve:
@tihuan Shouldn't this have typed the files?

seve:
@tihuan I thought our stance going forward with TS was that any changed files would also check in typings for the files at some level.  In this case at the least `renderContributors` and `client/src/components/util/version.ts` would be typed.

We're actually intoducing untyped code in this PR

tihuan:
Ahh I see what you mean now! Thanks for bringing this up 👍 
Given that the PR was written by a BE engineer and was time sensitive, I didn't want typing to block the merge. But I'll add typings in a follow up PR tomorrow!

Thank you!

tihuan:
Follow up PR: https://github.com/chanzuckerberg/single-cell-explorer/pull/81 Thank you!

PR #75, opened 2021-09-21
fix: disable portion of noisy env screenshots

seve:


tihuan (approved):
Thanks for the PR, Seve!

Ah maybe comment them out, so we can use them for debugging flaky tests in the future?

Thank you!

Issue #70, opened 2021-09-16
Add Portal's collection index URL to config response


MillenniumFalconMechanic:
#### Need
The Portal's collection index URL is required to generate links from the Explorer to the Portal.

#### Approach
Add the Portal's collection index URL to the config response.

**Alternate Approach**
We also considered using `location.origin` (see example below) but this would only work where the Portal and Explorer share the same origin. 

```
window.CELLXGENE = {};
window.CELLXGENE.API = {
  prefix: `${location.origin}${location.pathname}api/`,
  version: "v0.2/",
};
window.CELLXGENE.PORTAL = {
  url: `${location.origin}`,
}
```

#### Definition of Done
Portal's collection index URL is returned from `/config`.

PR #83, opened 2021-09-23
feat: 70 add portal collection homepage to /config

ebezzi:
#### Reviewers
**Functional:** 
@MDunitz @MillenniumFalconMechanic 

**Readability:** 
@seve 

Ticket [#70](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/70)

---

## Changes
- Adds the portal home page to the `/config` endpoint


server/tests/unit/common/test_api.py
```diff
@@ -843,6 +841,43 @@ def test_dataset_metadata_api_fails_gracefully_on_connection_failure(self, mock_
 
         self.assertEqual(result.status_code, HTTPStatus.BAD_REQUEST)
 
+class TestConfigEndpoint(BaseTest):
```
ebezzi, line 844: Note that the tests in the other suites aren't independent, so adding this to an existing suite doesn't work (it causes other tests to fail). Rewriting this file to have independent tests would be great.

MDunitz: Are you referring to the server config setup? or something else that prevents the test from being independent?

MDunitz (approved):
lgtm

PR #93, opened 2021-09-28
feat: 88 add dataset_id to the metadata endpoint

ebezzi:
#### Reviewers
**Functional:** 
@seve @MillenniumFalconMechanic 

---

## Changes
- Adds the `dataset_id` to the metadata endpoint


MillenniumFalconMechanic (approved):
LGTM, thank you @ebezzi!

MDunitz (approved):
Just checking what are we using the dataset id for? if it is to link to the explorer we should use the explorer_url instead bc there are datasets that dont use the id in the url

ebezzi:
> Just checking what are we using the dataset id for? if it is to link to the explorer we should use the explorer_url instead bc there are datasets that dont use the id in the url

@MillenniumFalconMechanic do you know? Madison has a good point.

MillenniumFalconMechanic:
We need a mechanism to filter out the current dataset from `collection_datasets` array so we can display only the sibling datasets in the dropdown menu.

PR #96, opened 2021-09-29
Fix flaky EB test

blrnw3:
It fails if 'make pydist' is not run, which installs the server code. This change allows it to run without the server code being installed.

TODO: move EB tests outside of the unit tests (story to come)


metakuni (approved):
🚀

PR #112, opened 2021-10-05
Dan/remove cli 2

danieljhegeman:
I decided to re-introduce actual cli code for the purpose of easily allowed any/all cmd line options to be used when doing local adhoc/dev work. Otherwise devs would have been limited compared to past state of cellxgene. New PR was necessitated in order to add the relevant code back in while also maintaining author history for it (so that it wouldn't appear to be added by me).

No major functional changes added after the first PR beyond reintroduction of cmd line option parsing, so give it a quick once-over. My apologies for the annoyance.

#### Reviewers
**Functional:** 
@MDunitz 
**Readability:** 
@blrnw3 
---

## Changes
- remove usage of cellxgene as a package
- run tests with multi dataset paradigm
- Separated out TODO comments on Flask routing into new Zenhub issue: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/110


common.mk
```diff
@@ -24,7 +24,7 @@ export JEST_ENV := $(call env_or_else_default,JEST_ENV)
 
 .PHONY: start-server
 start-server:
-	cellxgene launch -p $(CXG_SERVER_PORT) $(CXG_OPTIONS) $(DATASET)
+	PROJECT_ROOT=$(PROJECT_ROOT) python -m server.cli.launch
```
blrnw3, line 27: Shouldn't this call the `launch_server.sh` script and pass through all the args?
Or is this tricky with make files... it might be. Wondering if we should just remove this make target and use the script if so.

blrnw3 (approved):
Looks good afaict

common.mk
blrnw3: ok cool, that makes sense

MDunitz (approved):
lgtm

setup.cfg
```diff
@@ -1,4 +0,0 @@
-[flake8]
-max-line-length = 120
```
MDunitz, line 2: I think this removal causing your issues with the linter

danieljhegeman: good catch

server/tests/unit/__init__.py
```diff
@@ -149,6 +151,26 @@ def _before_adding_routes(app, app_config):
         if app_config.server_config.app__debug:
             CORS(app, supports_credentials=True)
 
+    @staticmethod
+    def _extract_base_url_and_dataset_for_api_calls(app_config: AppConfig):
```
danieljhegeman, line 155: convenience method for generating the url for the developer to use to access the front end since it is dependent on the name of the dataset being used (and the `CXG_CLIENT_PORT`, if set)

PR #116, opened 2021-10-06
Fix #104 dataframe histogram memoization

### LINKED ISSUE #104: Categorical variable slow to expand
Steps to reproduce:

1. Go tot tabula sapiens on explorer: https://cellxgene.cziscience.com/e/53d208b0-2cfd-4366-9866-c3c6114081bc.cxg/
2. Add gene CLDN7 in Quick Gene Search
3. Color by CLDN7
4. Click on the categorical variable cell type to expand

Cell type takes ~4 sec to expand. We should reduce this time or add a spinner of some sort.

blrnw3:
Memo was being set up on every call rather than once per instance, causing some UI features to be slow
e.g. https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/104.

These were the only instances of memo not working correctly that I could find. Also, this issue does not affect cellxgene desktop as the bug was introduced during the TS migration.

Perf screenshots when clicking a CatVar drop down (following the steps to reproduce above).
Before
Screen Shot 2021-10-06 at 9 55 37 AM

After
Screen Shot 2021-10-06 at 9 53 51 AM

#### Reviewers
**Functional:** 
Bruce
**Readability:** 
Seve
---

## Changes
* Fix memoization for dataframe's histogram functions

### DIFF
--- a/client/src/util/dataframe/dataframe.ts
+++ b/client/src/util/dataframe/dataframe.ts
@@ -251,6 +251,22 @@ class Dataframe {
     const { length } = column;
     const __id = __getMemoId();
     const isContinuous = isTypedArray(column);
+    const memoHistogramContinuous = memoize(
+      _histogramContinuous,
+      hashContinuous
+    );
+    const memoHistogramContinuousBy = memoize(
+      _histogramContinuousBy,
+      hashContinuousBy
+    );
+    const memoHistogramCategorical = memoize(
+      _histogramCategorical,
+      hashCategorical
+    );
+    const memoHistogramCategoricalBy = memoize(
+      _histogramCategoricalBy,
+      hashCategoricalBy
+    );
 
     /* get value by row label */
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):
Nice fix, which looks good.  And embarrassing given this was my bug!

It is very likely the cellxgene repo has the same issue.  Can you apply the fix to both?

blrnw3:
> It is very likely the cellxgene repo has the same issue. Can you apply the fix to both?

@bkmartinjr It looks like cellxgene is not affected because the bug was introduced during the TS migration, which seems to have only been done on the hosted code. Is there a plan to convert cellxgene to TS?

bkmartinjr:
> Is there a plan to convert cellxgene to TS?

No, there is not.  Thanks for checking!

Issue #126, opened 2021-10-07
Hide dataset selector on narrow viewports


MillenniumFalconMechanic:
#### Definition of Done
* Dataset selector is hidden when there is not enough width to display the truncating breadcrumbs in its most compact form.

PR #128, opened 2021-10-07
Added hide of selector on narrow viewports. #126

MillenniumFalconMechanic:
#### Reviewers
**Functional:** 
@tihuan, @frano-m 

---

## Changes
* Added HIDDEN, HIDDEN, HIDDEN state to state transitions enabling hide of dataset selector once it is no longer able to fit within available width.


tihuan (approved):
LGTMMM!!

Issue #89, opened 2021-09-28
Reenable dataset drawer 


MillenniumFalconMechanic:
#### Need
* Dataset drawer is to be re-enabled once schema updates are complete.

#### Approach
* Reattach click handler to dataset name.
* Update dataset drawer to handle schema updates.
* Leave dataset name and tooltip unchanged.

#### DoD
* Dataset name is displayed in logo area.
* Pointer cursor is displayed on hover on dataset name.
* Tooltip is displayed on hover on dataset name.
* Clicking dataset name results in dataset drawer opening.

PR #130, opened 2021-10-08
Removed data portal props from drawer. (#89)

MillenniumFalconMechanic:
#### Reviewers
**Functional:** 
@tihuan, @seve 

---

## Changes
* Removed data portal props prop from drawer as organism is now accessed via obs.
* Fixed link sort order.

### DIFF
--- a/client/src/components/infoDrawer/infoDrawer.tsx
+++ b/client/src/components/infoDrawer/infoDrawer.tsx
@@ -7,7 +7,7 @@ import { Drawer, Position } from "@blueprintjs/core";
 import InfoFormat, { SingleValueCategories } from "./infoFormat";
 import { AppDispatch, RootState } from "../../reducers";
 import { selectableCategoryNames } from "../../util/stateManager/controlsHelpers";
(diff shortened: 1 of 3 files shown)


client/src/components/infoDrawer/infoDrawer.tsx
```diff
@@ -41,7 +40,6 @@ type Props = DispatchProps & OwnProps & StateProps;
  */
 const mapStateToProps = (state: RootState): StateProps => ({
   datasetMetadata: state.datasetMetadata?.datasetMetadata,
-  dataPortalProps: state.config?.corpora_props,
```
seve, line 44: We can probably delete the reducer too, if it's not used anywhere else

seve (changes requested):

client/src/components/infoDrawer/infoDrawer.tsx
seve: That's right! I misread the reducer name and thought it was state.corpora_props

seve (approved):

Issue #141, opened 2021-10-12
Do not include datasets larger than 2 million cells in the dataset menu


brianraymor:
Datasets with more than 2 million cells (as documented by `cell_count` in the dataset-metadata response) MUST be filtered and not included in the dataset menu in _seamless_. 

An example dataset is the 4 million cell _Survey of human embryonic development_ in the [**A human cell atlas of fetal gene expression**](https://cellxgene.cziscience.com/collections/c114c20f-1ef4-49a5-9c2e-d965787fb90c). In this case, only the smaller _Survey of human embryonic development (1 million cells subset)_ would be included in the menu.

PR #142, opened 2021-10-13
Filtered out large datasets. (#141)

MillenniumFalconMechanic:
#### Reviewers
**Functional:** 
@tihuan, @ebezzi 

---

## Changes
- Filtered out large datasets (where cell count is greater than 2 million) on response from `/dataset-metadata`.

### DIFF
--- a/client/src/globals.ts
+++ b/client/src/globals.ts
@@ -9,6 +9,9 @@ export const overflowCategoryLabel = ": all other labels";
 /* default "unassigned" value for user-created categorical metadata */
 export const unassignedCategoryLabel = "unassigned";
 
+/** Maximum number of cells a dataset can have in order to be included for display. */
+export const DATASET_MAX_CELL_COUNT = 200000;
+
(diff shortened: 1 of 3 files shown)


client/src/globals.ts
```diff
@@ -9,6 +9,9 @@ export const overflowCategoryLabel = ": all other labels";
 /* default "unassigned" value for user-created categorical metadata */
 export const unassignedCategoryLabel = "unassigned";
 
+/** Maximum number of cells a dataset can have in order to be included for display. */
+export const DATASET_MAX_CELL_COUNT = 200000;
```
tihuan, line 13: Two million is the max 🙆

```suggestion
export const DATASET_MAX_CELL_COUNT = 2_000_000;
```

tihuan (approved):
Looks really good! Thanks so much for the PR, Mim! Just a small bit for changing the max to two million 😸 Thank you!!

PR #159, opened 2021-10-19
fix: 156 obsolete browsers

tihuan:
#### Reviewers
**Functional:** 
@seve
**Readability:** 
@dimple-jethani 
---

## Changes
This PR replaces `obsolete-webpack-plugin` and uses [`browserslist-useragent-regexp`](https://github.com/browserslist/browserslist-useragent-regexp#readme) to generate a REGEX to test the `navigator.userAgent` instead

The strategy is:

1. Use `browserslist-useragent-regexp` to generate REGEX based on `browserslist` and output a file `configuration/webpack/SUPPORTED_BROWSERS_REGEX.js`
2. `configuration/webpack/webpack.config.shared.js` imports the REGEX and obsolete html template and inject them into both `index.html` and `index_template.html` on the designated script tag.
3. `configuration/webpack/webpack.config.shared.js` also copies the static file `configuration/webpack/obsoleteBrowsers.js` into the `/build/static/` directory, so the servers know how to handle the request for the file
4. `index.html` and `index_template.html` then loads the script file script tag in step 2 via `src="obsoleteBrowsers.js"`. NOTE: we continue to use relative path here instead of absolute path, so the request will still be prefixed with data root path (the namespace needed for CloudFront)
5. `obsoleteBrowsers.js` is now downloaded asynchronously and parse REGEX and HTML template from its script tag attributes and test the browser and render the obsolete HTML as needed

Safari:

Screen Shot 2021-10-19 at 3 30 38 AM

Chrome:

Screen Shot 2021-10-19 at 3 31 26 AM

🚨🚨🚨 TODO 🚨🚨🚨
~1. Still need to remove associated script hash in `single-cell-infra` repo~ False alarm

### DIFF
--- a/client/src/common/types/entities.ts
+++ b/client/src/common/types/entities.ts
@@ -1,11 +1,6 @@
 // If a globally shared type or interface doesn't have a clear owner, put it here
 
 /**
- * Category name suffix used to determine if category name is ontology term id.
- */
-export const ONTOLOGY_KEY = "ontology_term_id";
-
-/**
  * Complete list of standard cellxgene column names.
  * https://github.com/chanzuckerberg/single-cell-curation/blob/main/schema/2.0.0/schema.md#obs-cell-metadata
  */
--- a/client/src/components/categorical/index.tsx
+++ b/client/src/components/categorical/index.tsx
@@ -3,10 +3,7 @@ import { AnchorButton, Tooltip, Position } from "@blueprintjs/core";
 import { connect } from "react-redux";
 import * as globals from "../../globals";
 import Category from "./category";
-import {
-  ONTOLOGY_KEY,
-  STANDARD_CATEGORY_NAMES,
-} from "../../common/types/entities";
+import { STANDARD_CATEGORY_NAMES } from "../../common/types/entities";
 import Collapse from "../../util/collapse";
 import { AnnotationsHelpers, ControlsHelpers } from "../../util/stateManager";
 import AnnoDialog from "../annoDialog";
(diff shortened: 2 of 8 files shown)


client/configuration/webpack/obsoleteHTMLTemplate.html
```diff
@@ -1,12 +1,3 @@
-<script>
-  /** Reflect ANY changes to this script in the script hash in `server/eb/app.py` **/
-  var root = document.getElementById("root");
-  root.remove();
-  var portals = document.getElementsByClassName("bp3-portal");
-  for (var i = 0; i < portals.length; i += 1) {
-    portals[i].remove();
-  }
-</script>
```
seve, line 9: We're going to need to update CSP

client/index.html
```diff
@@ -42,5 +42,12 @@
     >
 
     <div id="root"></div>
+    <script
+      async
+      src="static/obsoleteBrowsers.js"
+      id="obsolete-browsers"
+      data-template="<%= OBSOLETE_TEMPLATE %>"
+      data-regex="<%= OBSOLETE_REGEX %>"
+    ></script>
```
seve, line 51: Is this going to need to be reflected in CSP?

client/configuration/webpack/obsoleteBrowsers.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable @blueprintjs/classes-constants -- do not import files */
```
seve, line 1: Maybe this too will need to be included in CSP?

Issue #147, opened 2021-10-15
The hover for a singleton in the dataset drawer should not be the ontology identifier


brianraymor:
We eliminated the ontology identifiers from the LSB categorical hovers. The singletons in the dataset drawer should be consistent.

PR #163, opened 2021-10-20
Removed ontology ID tooltip from drawer. (#147)

MillenniumFalconMechanic:
#### Reviewers
**Functional:** 

---

## Changes
- Removed ontology tooltip from drawer.


tihuan (approved):
Ahh makes sense! Thanks for the refactoring, Mim 🎉 !

Issue #183, opened 2021-10-20
Remove matrix_data_cache_manager


MDunitz:
## Need
Remove matrix_data_cache_manager. With Cloudfront providing matrix data caching, Explorer BE should no longer need to provide its own server-side caching. Also, the backend server can retrieve the TileDB CXG data performantly from S3, so caching is much less important for that use case (we do not send these raw data across the network to the client).

## Approach 
Once [167](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/167) is merged remove code related to matrix_data_cache_manager from the single-cell-explorer code base and from the configs in single-cell-infra

## DOD 
- response time for endpoints is re-benchmarked (as described in [165](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/165))

PR #187, opened 2021-11-02
rm matrix cache mgr [#183]

atolopko-czi:
#### Reviewers
**Functional:** 
@Bento007 

**Readability:** 
@danieljhegeman 

---

## Changes

The matrix data will be cached by CDN and user's browser so no need to cache ourselves at the server level. Cached items will be keyed by the S3 URL, so the data is immutable and can exist in CDN/browser caches for eternity.

- Remove all references to matrix data cache
- Remove caching config
- Remove cache classes, no longer used anywhere


server/data_common/data_adaptor.py
```diff
@@ -122,6 +122,12 @@ def get_var_keys(self):
         # return list of keys
         pass
 
+    def __enter__(self):
+        return self
```
danieljhegeman, line 126: this was a good little learning for me -- for some reason I always thought that `__enter__` had a default definition that returned self and that `__exit__` had a default definition that did nothing.

But apparently you had to define these, as I just learned from a little test on my own.

TIL 👍

server/default_config.py
```diff
@@ -76,15 +76,6 @@
...
-    matrix_cache:
-      # The maximum number of datasets that may be opened at one time.  The least recently used dataset
-      # is evicted from the cache first.
-      max_datasets: 5
-
-      # A matrix is automatically removed from the cache after timelimit_s number of seconds.
-      # If timelimit_s is set to None, then there is no time limit.
-      timelimit_s: 30
-
   single_dataset:
```
danieljhegeman, line 79: we can convene later on removing everything related to single dataset mode

danieljhegeman: good cleanup task

server/data_common/data_adaptor.py
atolopko-czi: yeah, also still adjusting to not having explicit interfaces and/or base classes for all the special methods Python supports.  https://en.wikipedia.org/wiki/Duck_typing

server/default_config.py
atolopko-czi: I gotcha covered: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/181

Bento007 (approved):
I love commits where it's more deletions than additions. LGTM

danieljhegeman:
tests pass 👍

Bento007:
Do these changes depend on any remaining external work?

danieljhegeman (approved):

atolopko-czi:
> Do these changes depend on any remaining external work?

Can you clarify "external work"? But in case I understand, we do need to deploy this along with v0.3 API changes, since removal of the caches relies upon the having API callers supply the dataset artifact S3 URLs.

PR #188, opened 2021-11-02
rdev build image only

atolopko-czi:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/255

### Reviewers
**Functional:** 
@ebezzi 

**Readability:** 
@metakuni 

---

## Changes
Pushing a branch or tag with an `rdev` prefix or suffix will now build & push (to ECR) the Docker image, but will NOT deploy to rdev. The deploy step is now performed exclusively with the `single-cell-infra` repo's `happy` command, which creates/updates an rdev stack using a specific docker image tag that must exist on both explorer and portal images.


.happy/config.json
```diff
@@ -0,0 +1,14 @@
+{
+    "config_version": "v1",
+    "terraform_version": "0.13.5",
+    "default_env": "rdev",
+    "app": "hosted-cellxgene",
+    "default_compose_env": ".env.ecr",
+    "environments": {
```
metakuni, line 7: We don't need to define environments for dev, stage, and prod?

metakuni (approved):
Next up: mono repo to consolidate 😄

.happy/config.json
atolopko-czi: All happy config in explorer is currently for rdev purposes only.

.happy/config.json
```diff
@@ -0,0 +1,14 @@
...
+        "rdev": {
```
metakuni, line 8: Do we not need `terraform_directory` here? (just comparing to `config.json` for portal)

.happy/config.json
```diff
@@ -0,0 +1,14 @@
...
+            "aws_profile": "single-cell-dev",
+            "secret_arn": "happy/env-rdev-config",
+            "log_group_prefix": "/dp/rdev"
```
metakuni, line 11: Should this be something like `/hc/rdev` (hosted cellxgene) instead?

.happy/config.json
metakuni: Ah, nm, that's in the infra repo. 🤦

atolopko-czi: Right, `happy` script doesn't need terraform stuff here since we're just calling `push` subcmd to build the docker image.  I really don't like having any `happy` stuff in this repo at all, and we can eliminate all this with an alternate rdev deployment design: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/253#issuecomment-967169708

.happy/config.json
atolopko-czi: yes, though I'm not sure this is even being used here, since it's infrastructure-related. Unless it's getting baked into the docker image.  Ideally, we can just delete this line if it's not used.

ebezzi (approved):

PR #192, opened 2021-11-08
cache redesign epic

atolopko-czi:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
Remove server caching:
* Remove dataset metadata cache. The metadata will only be requested once by the Explorer client now. And each new Explorer session _should_ request fresh metadata from Portal, so caching no longer desirable.
* Remove matrix data cache. The matrix data will be cached by CDN and user's browser so no need to cache ourselves at the server level. Cached items will be keyed by the S3 URL, so the data is immutable and can exist in CDN/browser caches for eternity.
* Remove base cache classes, no longer used anywhere
* Remove caching config

v3 API changes:
* Add v3 API endpoints, splitting v2 and v3 endpoints into separate modules. v3 endpoints use dataset artifact S3 URI to retrieve CXG array data.
* Adding S3 URI API endpoint for client to retrieve the value that will be need to make requests to v3 dataset endpoints
* Drop support for dataroot-specific dataset configs. These can no longer be supported in v3 api, due to lack of dataroot in api requests, but were not being used.
* Use consistent dataset_root and dataset_id params to identify the Explorer location of a dataset.* add s3 fetch step to modify api url

FE changes:
* update API version to v0.3
* request dataset artifact S3 URI at init time, and use in subsequent v0.3 dataset endpoint requests
* change return shape of uri
* rework metadata api and begin typing
* Update client/src/globals.ts


server/app/api/__init__.py
```diff
@@ -0,0 +1,40 @@
+from functools import wraps
```
atolopko-czi, line 1: TODO: Should move out of `__init__.py`

Bento007 (approved):

Issue #136, opened 2021-10-11
Hide super categories if categories is empty


MillenniumFalconMechanic:
#### Approach
* Match existing functionality for Author super category.

#### Definition of Done
* Standard super category is hidden if it has no categories.
* Continuous super category is hidden if it is empty.

brianraymor:
I've found a case where this fails. If you watch the video, it looks like batchA and batchB singletons appear and immediately disappear in the Author super-category and then the Author category remains. The singletons show up in the dataset drawer.

https://images.zenhubusercontent.com/397736859/a44b8731-c561-4460-ab73-7014c1ec1b21/screen_recording_2021_11_03_at_5_25_26_pm.mov

PR #193, opened 2021-11-09
Updated author supercategory partitioning. (#136)

MillenniumFalconMechanic:
#### Reviewers
**Functional:** 
@tihuan, @seve 

---

## Changes
- Updated calculations of standard and author categories for display in the LSB.

### DIFF
--- a/client/src/components/categorical/index.tsx
+++ b/client/src/components/categorical/index.tsx
@@ -133,6 +133,41 @@ class Categories extends React.Component<{}, State> {
     }
   };
 
+  /**
+   * Determine if category name is an ontology key.
+   * @param catName - Name of category.
+   * @returns True if category name includes ontology key.
+   */
+  isCategoryNameOntologyKey = (catName: string): boolean =>
+    catName.includes(globals.ONTOLOGY_KEY);
+
+  /**
+   * Categories are included for display if category has more than one category value or categories are not defined
+   * (for the case where category is a string or boolean type).
+   * @param schema - Matrix schema.
+   * @param catName - Name of category.
+   * @returns True if category has more than one category value or categories are not defined.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types -- update on typing of component
+  isCategoryDisplayable = (schema: any, catName: string): boolean =>
+    schema.annotations.obsByName[catName].categories?.length > 1 ||
+    !schema.annotations.obsByName[catName].categories;
+
+  /**
+   * Determine if category is standard.
+   * @param catName - Name of category.
+   * @returns True if given category name is in the set of standard category names.
+   */
+  isCategoryNameStandard = (catName: string): boolean =>
+    STANDARD_CATEGORY_NAMES.includes(catName);
+
+  /**
+   * Returns true if category is writable.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types -- update on typing of component
+  isCategoryWritable = (schema: any, catName: string): boolean =>
+    schema.annotations.obsByName[catName].writable;
+
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
   render() {
     const {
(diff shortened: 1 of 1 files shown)


client/src/components/categorical/index.tsx
```diff
@@ -133,6 +133,41 @@ class Categories extends React.Component<{}, State> {
...
+    catName.includes(globals.ONTOLOGY_KEY);
+
+  /**
+   * Categories are included for display if category has more than one category value or categories are not defined
+   * (for the case where category is a string or boolean type).
+   * @param schema - Matrix schema.
+   * @param catName - Name of category.
+   * @returns True if category has more than one category value or categories are not defined.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types -- update on typing of component
```
seve, line 151: Maybe we should adopt a standard for these types of comments to make it easier to undo the disables once the appropriate component(s) are typed.  cc: @tihuan

client/src/components/categorical/index.tsx
```diff
@@ -133,6 +133,41 @@ class Categories extends React.Component<{}, State> {
...
+  isCategoryDisplayable = (schema: any, catName: string): boolean =>
+    schema.annotations.obsByName[catName].categories?.length > 1 ||
+    !schema.annotations.obsByName[catName].categories;
```
seve, line 154: We can probably change this line since we have typing to make it more readable!

Maybe something like 
```suggestion
    !schema.annotations.obsByName[catName] instanceof CategoricalAnnotationColumnSchema
```

tihuan (commented):
LGTM! Probably wait for @seve , since he knows this code much better than I do 😆 Thank you!

seve (approved):
Overall approval, the TSDoc strings are beautiful 🤩 

Let's resolve my comments before merge!

client/src/components/categorical/index.tsx
seve: It should be something easily cmd + f -able.  Either an issue # or the file path that needs to be typed

tihuan: Oh great question 💡 I think special comment type would be useful if we do `schema: any` because `schema` is not typed yet. In this case, I think `schema` is already typed elsewhere, so it's just a matter of using when we type everything in this file?

Also I'm wondering maybe `@typescript-eslint/explicit-module-boundary-types` is not needed, since we have `(...): boolean => `?

MillenniumFalconMechanic: I have update the `schema` argument type in both [isCategoryDisplayable](https://github.com/chanzuckerberg/single-cell-explorer/pull/193/files#diff-7f6832c58dab222ca673a6841ad0efb0f300c20eee501bee503fe3d40f0c1b44R152) and [isCategoryWritable](https://github.com/chanzuckerberg/single-cell-explorer/pull/193/files#diff-7f6832c58dab222ca673a6841ad0efb0f300c20eee501bee503fe3d40f0c1b44R176) to `Schema`.

Timmy, good q about `@typescript-eslint/explicit-module-boundary-types`. According to typescript-eslint's [supported rules](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules), it looks both the return type *and* argument types are required (in our case, on public methods):

```
Require explicit return and argument types on exported functions' and classes' public class methods
```

client/src/components/categorical/index.tsx
MillenniumFalconMechanic: I am getting a `TS2693: 'CategoricalAnnotationColumnSchema' only refers to a type, but is being used as a value here.` error using `instanceof`.

I have updated the code to use the `in` operator to narrow the type of `schema.annotations.obsByName[catName]` to `CategoricalAnnotationColumnSchema` when checking the `categories` length but I had trouble getting it any clearer than that. I'm open to suggestions on improving readability here!

client/src/components/categorical/index.tsx
tihuan: Ahh TIL! Thanks so much for sharing, @MillenniumFalconMechanic 🙏 SGTM!

client/src/components/categorical/index.tsx
tihuan: Hmm yeah I think if we really wanted to go one step further, we'd need to create a helper function like: 

```ts
function isCategoricalAnnotationColumnSchema(schema: AnnotationColumnSchema): schema is CategoricalAnnotationColumnSchema {
  if (("categories" in columnSchema)) {
      return true;
    }

  return false;
}
```

But I don't feel strongly about it, since TS already narrows the type like Mim says 👌 

And it makes sense that `!schema.annotations.obsByName[catName] instanceof CategoricalAnnotationColumnSchema` doesn't work, since `instanceof` only works on `classes` and `CategoricalAnnotationColumnSchema` is an interface?

tihuan (approved):
LGTM! Thanks for updating the PR, @MillenniumFalconMechanic 🙏 🔥

seve (approved):

PR #195, opened 2021-11-23
feat: client uses v0.3 api

seve:
#### Reviewers
**Functional:** @tihuan 

**Readability:** @Bento007, @MillenniumFalconMechanic 

---

## Changes
- add
   - `s3URIFetch`
   - `updateAPIWithS3()` swaps out preloaded API embedded in HTML with S3URI prefix
- remove
- modify
   - Modify templates to use `v0.3`
   - Changed naming of some endpoint request helper functions to better reflect fetch vs fetchAndLoad
   - Various typing and type safety

### DIFF
--- a/client/src/actions/annotation.ts
+++ b/client/src/actions/annotation.ts
@@ -476,9 +476,14 @@ export const saveGenesetsAction =
         }),
       ]);
     } catch (error) {
+      if (error instanceof Error)
+        return dispatch({
+          type: "autosave: genesets error",
+          message: error.toString(),
+          error,
+        });
       return dispatch({
         type: "autosave: genesets error",
-        message: (error as Error).toString(),
         error,
       });
     }
--- a/client/src/actions/annotation.ts
+++ b/client/src/actions/annotation.ts
@@ -320,6 +320,7 @@ export const needToSaveObsAnnotations = (
 export const saveObsAnnotationsAction =
   () =>
   async (dispatch: AppDispatch, getState: GetState): Promise<void> => {
+    if (!globals.API) throw new Error("API not set");
     const state = getState();
     const { annotations, autosave } = state;
     const { dataCollectionNameIsReadOnly, dataCollectionName } = annotations;
@@ -377,17 +378,19 @@ export const saveObsAnnotationsAction =
         });
       }
     } catch (error) {
-      dispatch({
-        type: "writable obs annotations - save error",
-        message: error.toString(),
-        error,
-      });
+      if (error instanceof Error)
+        dispatch({
+          type: "writable obs annotations - save error",
+          message: error.toString(),
+          error,
+        });
     }
   };
 
 export const saveGenesetsAction =
   () =>
   async (dispatch: AppDispatch, getState: GetState): Promise<unknown> => {
+    if (!globals.API) throw new Error("API not set");
     const state = getState();
 
     // bail if gene sets not available, or in readonly mode.
@@ -475,7 +478,7 @@ export const saveGenesetsAction =
     } catch (error) {
       return dispatch({
         type: "autosave: genesets error",
-        message: error.toString(),
+        message: (error as Error).toString(),
         error,
       });
     }
--- a/client/src/actions/index.ts
+++ b/client/src/actions/index.ts
@@ -16,7 +16,7 @@ import * as genesetActions from "./geneset";
 import { AppDispatch, GetState } from "../reducers";
 import { EmbeddingSchema, Schema } from "../common/types/schema";
 import { ConvertedUserColors } from "../reducers/colors";
-import { DatasetMetadata, Dataset } from "../common/types/entities";
+import type { DatasetMetadata, Dataset, S3URI } from "../common/types/entities";
 import { postExplainNewTab } from "../components/framework/toasters";
 import { KEYS } from "../components/util/localStorage";
 import {
@@ -50,6 +50,10 @@ async function userColorsFetchAndLoad(
       userColors: loadUserColorConfig(response),
     })
   );
+}
+
+async function s3URIFetch(): Promise<{ s3_uri: S3URI }> {
+  return fetchJson<{ s3_uri: S3URI }>("s3_uri");
 }
 
 async function schemaFetch(): Promise<{ schema: Schema }> {
@@ -145,6 +149,8 @@ const doInitialDataLoad = (): ((
     dispatch({ type: "initial data load start" });
 
     try {
+      const { s3_uri: s3URI } = await s3URIFetch();
+      globals.updateAPIWithS3(s3URI);
       const [config, schema] = await Promise.all([
         configFetch(dispatch),
         schemaFetch(),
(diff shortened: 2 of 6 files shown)


client/src/actions/index.ts
```diff
@@ -72,15 +78,17 @@ async function configFetch(dispatch: AppDispatch): Promise<Config> {
 /**
  * Fetch dataset metadata and dispatch save to store, including portal URL returned in /config.
  * @param dispatch Function facilitating update of store.
+ * @param oldPrefix API prefix with dataset path that dataset metadata lives on. (Not S3 URI)
  * @param config Response from config endpoint containing collection ID for the current dataset.
  */
-async function datasetMetadataFetch(
+async function datasetMetadataFetchAndLoad(
   dispatch: AppDispatch,
+  oldPrefix: string,
```
seve, line 86: We're passing the old prefix (data root path) now, as this API call occurs later down the line and by this point the global has been overwritten with the S3URI prefix. (metadata lives on the data root)

client/src/actions/index.ts
```diff
@@ -222,18 +240,12 @@ const dispatchDiffExpErrors = (dispatch: any, response: any) => {
...
-    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  ) =>
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  async (dispatch: any, getState: any) => {
+  (set1: LabelArray, set2: LabelArray, num_genes = 50) =>
+  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types -- TODO: type diff exp data in genesets reducer (client/src/reducers/genesets.ts)
+  async (dispatch: AppDispatch, getState: GetState) => {
     dispatch({ type: "request differential expression started" });
     try {
+      if (!globals.API) throw new Error("API not set");
```
seve, line 248: Is there a cleaner way of throwing this error? Realistically, if there is no API set, the app is DOA and we have already thrown an error

client/src/globals.ts
```diff
@@ -177,3 +188,19 @@ export function updateApiPrefix(): void {
...
+ * @param s3Uri S3 URI of the current dataset.
+ * @returns The old API prefix.
+ */
+export function updateAPIWithS3(s3URI: S3URI): string {
+  if (!API) {
+    throw new Error("Unable to set API route.");
+  }
+  const oldAPI = API.prefix;
+  const URISafeS3URI = encodeURIComponent(s3URI);
+  const flaskSafeS3URI = `s3_uri/${encodeURIComponent(URISafeS3URI)}/`;
```
seve, line 203: Needs to be double encoded

Bento007 (commented):
It looks good to me, but I'd wait for Timmy's review.

client/src/actions/index.ts
atolopko-czi: Since this var name assumes prior knowledge of what "old" means, maybe it can be named more explicitly for future engineers? 

E.g., in the BE code, it's being called "[dataset] Explorer URL path".  

(Though BE should probably change "[dataset] Explorer URL path" to "... URI ...", to be correct, as it does not include an HTTP scheme in the path.)

atolopko-czi: > by this point the global has been overwritten with the S3URI prefix

What if we just main two prefixes on `API`, one for the "public" dataset identifier (i.e. the "explorer url") and one for the "internal" dataset _artifact_ identifier? It's conceivable we'll add more API calls in the future that need the former.  

I think we should also settle on some clear variable names for:
* the "public" dataset identifier (i.e. the "explorer url"): `datasetPublicPath` or `datasetPublicIdentifier`
* the "internal" dataset _artifact_ identifier: e.g. `datasetArtifactUri`

And then use these throughout the BE and FE code:
* `s3URIFetch()` -> `datasetArtifactUriFetch()`
* `updateAPIWithS3()` -> `updateAPIWithDatasetArtifactUri()`

atolopko-czi: Maybe a quick meeting today to align on naming? If/when naming changes are decided upon, we can address changes as a new refactoring story.

seve: No strong preference on naming. For some context, the word `old` has a runtime connotation. Old as in before it was rewritten ms ago.

We could do something a bit more literal and how it's written in my comments, such as:
 * s3BaseAPIPrefix - but once this is used as the actual APIPrefix I think the default API naming is proper
 * dataRootAPIPrefix

client/src/actions/index.ts
tihuan: Yeah I think this is fine, since it's unlikely we'll run into this condition IRL, but defensive programming is great. So keeping it sounds good!

client/src/globals.ts
```diff
@@ -151,23 +152,33 @@ const CXG_SERVER_PORT =
...
+    CELLXGENE: {
+      API: {
+        prefix: string;
+        version: string;
+      };
+    };
   }
+}
+
+if (window.CELLXGENE && window.CELLXGENE.API) {
```
tihuan, line 166:
```suggestion
if (window?.CELLXGENE?.API) {
```

client/src/globals.ts
tihuan: I think it's worth adding this as a comment and explains why it has to be double encoded 🙆‍♂️

tihuan (approved):
LGTM! Looks really robust 👍 Just some small nits and we can merge first and do the naming thing as a separate PR! Thank you 🎉

client/src/actions/index.ts
atolopko-czi: Thank you for weighing in on the naming! I'm going to add a new story to address naming changes on both FE and BE at same time, to keep them aligned. Will include current naming suggestions, and will seek broader team input as well.

atolopko-czi: https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/196

client/src/actions/annotation.ts
```diff
@@ -475,7 +478,7 @@ export const saveGenesetsAction =
     } catch (error) {
       return dispatch({
         type: "autosave: genesets error",
-        message: error.toString(),
+        message: (error as Error).toString(),
```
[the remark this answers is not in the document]
seve: Oops I actually wanted to swap the `instanceof` above out for the as Error.

What else can trigger the catch other than an Error?

client/src/globals.ts
seve:
```suggestion
  // must be double quoted so slashes are not decoded early by flask WSGI.
  const URISafeS3URI = encodeURIComponent(s3URI);
  const flaskSafeS3URI = `s3_uri/${encodeURIComponent(URISafeS3URI)}/`;
```

PR #198, opened 2021-11-29
fix: remove trailing slash from s3_uri

Bento007:
#### Reviewers
**Functional:**  @atolopko-czi 

**Readability:** @seve 

---

Remove trailing slash from the s3_uri that is used in the `/s3_uri/<s3_uri>/api/v0.3` requests, to ensure that all requests are properly utilizing the CloudFront cache. This woull change this:

`/cellxgene/s3_uri/s3%253A%252F%252Fhosted-cellxgene-dev%252Fpbmc3k.cxg%252F/api/v0.3/genesets`

to this

`/cellxgene/s3_uri/s3%253A%252F%252Fhosted-cellxgene-dev%252Fpbmc3k.cxg/api/v0.3/genesets`

## Changes
- remove the trailing slash from S3_URIs before returning in GET */s3_uri

### DIFF
--- a/server/app/api/util.py
+++ b/server/app/api/util.py
@@ -5,7 +5,8 @@ from server.data_common.matrix_loader import MatrixDataLoader
 
 
 def get_dataset_artifact_s3_uri(url_dataroot: str = None, dataset_id: str = None):
-    dataset_artifact_s3_uri = get_dataset_metadata(url_dataroot, dataset_id, current_app.app_config)["s3_uri"].rstrip('/')
+    dataset_artifact_s3_uri = get_dataset_metadata(url_dataroot, dataset_id, current_app.app_config)["s3_uri"]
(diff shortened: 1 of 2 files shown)


atolopko-czi (approved):
LGTM.  Could be worth adding a test to help protect any inadvertent removal of this no-slash requirement. Similarly adding a comment here to explain the purpose would be useful.

atolopko-czi:
Beyond the scope of this PR, it seems worth "fixing" this in data portal as well by having an explicit, validated format there (i.e. no trailing slash).  Similarly for explorer URLs,  since there exists a hack to query for explorer URLs that may or may not have a trailing slash.  I can create stories for this, but seeking input.

Bento007:
It seems like a low priority, but something worth doing eventually.

PR #205, opened 2021-12-13
fix: add max_age to immutable caches

seve:
#### Reviewers
**Functional:** @atolopko-czi 

**Readability:** @ebezzi


seve:
#2476

atolopko-czi (approved):
Thanks for jumping on this Seve! Approving optimistically, with rdev env browser test being the final arbiter of success.

I suppose we could DRY the `@cache_control(immutable=True, max_age=ONE_YEAR)` line into a new annotation like `@cache_immutable`, but maybe that's just obscuring the caching details unnecessarily.

PR #234, opened 2022-03-11
Differential expression performance improvements

blrnw3:
#### Reviewers

**Functional:**   @atolopko-czi  
**Readability:**   @ebezzi 

---

## Changes
* Remove legacy h5ad diffex algo
* Update to latest tiledb
* Optimize cxg diffex algo
   * remove parallelism logic (tdb handles this better) and config params for this code
   * tweak tdb queries based on Bruce's research
* Optimize other tdb queries (the get_X_array calls)
* Remove col shift logic - legacy
* Update tdb context config and related tests
* Update run_diffex script


server/compute/diffexp_cxg.py
```diff
@@ -1,120 +1,65 @@
...
+    :param maskB: observation selection mask for set 2
+    :param top_n: number of variables to return stats for
+    :param diffexp_lfc_cutoff: minimum
+    :param arr: the tdb array to read
+    absolute value returning [ varindex, logfoldchange, pval, pval_adj ] for top N genes
+    :return:  for top N genes, {"positive": for top N genes, [ varindex, foldchange, pval, pval_adj ],
+              "negative": for top N genes, [ varindex, foldchange, pval, pval_adj ]}
+    """
+    matrix = adaptor.open_array(arr)
     row_selector_A = np.where(maskA)[0]
```
bkmartinjr, line 45: Using `np.where` is slightly poor form and lacks clarity.  See numpy [docs](https://numpy.org/doc/stable/reference/generated/numpy.where.html):

> When only condition is provided, this function is a shorthand for np.asarray(condition).[nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html#numpy.nonzero)(). Using nonzero directly should be preferred, as it behaves correctly for subclasses.

bkmartinjr: Oh, and I apologize if this was originally my error!

server/dataset/cxg_dataset.py
```diff
@@ -23,7 +23,12 @@
 class CxgDataset(Dataset):
     # TODO:  The tiledb context parameters should be a configuration option
     tiledb_ctx = tiledb.Ctx(
-        {"sm.tile_cache_size": 8 * 1024 * 1024 * 1024, "sm.num_reader_threads": 32, "vfs.s3.region": "us-east-1"}
+        {
+            "sm.tile_cache_size": 16 * 1024 ** 3,
```
bkmartinjr, line 27: will these changes be accompanied with a change of instance type?  The large tile_cache_size and py.init_buffer_bytes are suitable for a high-mem instance, but not for the current 64GB instances.  Thought: should these be pulled out into a deployment configuration (eg, set via env vars), or made adaptive (eg, tile cache is X% of RAM on the instance)?

server/compute/diffexp_cxg.py
```diff
@@ -1,120 +1,65 @@
...
-    if diffexp_thread_executor is None:
-        diffexp_thread_executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
-    return diffexp_thread_executor
-
-
-def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01):
-
-    matrix = adaptor.open_array("X")
+def set_config(*args, **kwargs):
+    # TODO: find callers and eliminate calls
```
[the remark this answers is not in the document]
blrnw3: I tried, but there are indeed callers, didn't track them down but we should

server/compute/diffexp_cxg.py
```diff
@@ -165,35 +199,14 @@ def _mean_var_sparse_numba(x, var, nrows, ncols):
...
-    if fp_err_occurred:
-        mean[np.isfinite(mean) == False] = 0  # noqa: E712
-        v[np.isfinite(v) == False] = 0  # noqa: E712
-    else:
-        mean[np.isnan(mean)] = 0
-        v[np.isnan(v)] = 0
-
-    return mean, v
+def mean_var_cnt_dense(matrix, _, rows):
+    # TODO: why is the list cast still necessary? Tdb is throwing an error
```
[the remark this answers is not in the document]
blrnw3: Agreed, will add to the task list

server/compute/diffexp_cxg.py
blrnw3: This was the original code, not sure who wrote it but good point, we should update it!

server/dataset/cxg_dataset.py
blrnw3: Yes there's an infra PR here: https://github.com/chanzuckerberg/single-cell-infra/pull/526
Re: making them dynamic based on instance - agreed!

server/dataset/cxg_dataset.py
```diff
@@ -265,18 +270,12 @@ def get_X_array(self, obs_mask=None, var_mask=None):
             if obs_items == slice(None) and var_items == slice(None):
                 data = X[:, :]
```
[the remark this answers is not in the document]
blrnw3: Good spot. Though is this even used. TODO: check if it is

server/tests/unit/dataset/test_cxg_dataset.py
```diff
@@ -15,6 +20,15 @@ def test_get_colors(self):
         self.assertDictEqual(data.get_colors(), dict())
 
     def get_data(self, fixture):
+        # FIXTURES_ROOT = "/home/ec2-user/server/tests/fixtures"
         data_locator = f"{FIXTURES_ROOT}/{fixture}"
         config = app_config(data_locator)
         return CxgDataset(DataLocator(data_locator), config)
+
+    def test_tdb_bug(self):
+        """ This gives different results on 0.12.4 vs 0.13.1. TODO: fix """
```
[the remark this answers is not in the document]
bkmartinjr: resolved by work-around added to server/dataset/cxg_dataset.py:set_tiledb_context()

server/compute/diffexp_cxg.py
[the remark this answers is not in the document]
bkmartinjr: Update: this is a bug in recent TileDB-Py versions. In the queue for a fix in TileDB 0.13.2, at which point multi_index should again support either list or ndarray.

server/compute/diffexp_cxg.py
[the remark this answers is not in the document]
bkmartinjr: tracked down and removed.

server/compute/diffexp_cxg.py
bkmartinjr: modified to us numpy recommended style

server/dataset/cxg_dataset.py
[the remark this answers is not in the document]
bkmartinjr: It is used.  Added a commit to simplify the logic, as multi_index[] can provide both query types -- we don't need the branch.

bkmartinjr (approved):
I have resolved my own prior comments, but would like additional reviewers for the new modifications.

server/dataset/cxg_dataset.py
```diff
@@ -21,9 +21,13 @@
...
 
 class CxgDataset(Dataset):
-    # TODO:  The tiledb context parameters should be a configuration option
+    # These defaults are overridden by the config variable: server.adaptor.cxg_adaptor.tiledb_cxt
     tiledb_ctx = tiledb.Ctx(
-        {"sm.tile_cache_size": 8 * 1024 * 1024 * 1024, "sm.num_reader_threads": 32, "vfs.s3.region": "us-east-1"}
+        {
+            "sm.tile_cache_size": 8 * 1024 ** 3,
+            "py.init_buffer_bytes": 16 * 1024 ** 3,
+            "vfs.s3.region": "us-east-1",
```
ebezzi, line 29: I was wondering why the region is even required. Our buckets live in `us-west-2` so it looks like this is ignored by TileDB.

server/dataset/cxg_dataset.py
```diff
@@ -244,7 +240,7 @@ def __remap_indices(self, coord_range, coord_mask, coord_data):
         if coord_mask is None:
             return coord_range, coord_data
 
-        indices = np.where(coord_mask)[0]
+        indices = coord_mask.nonzero()[0]
```
ebezzi, line 243: Out of curiosity, is this more efficient than using `np.where`?

ebezzi (approved):

server/dataset/cxg_dataset.py
bkmartinjr: no.  They are literally identical, but np.where is deprecated in this case.  See the numpy documentation here: https://numpy.org/doc/stable/reference/generated/numpy.where.html

It says:
```
Note

When only condition is provided, this function is a shorthand for np.asarray(condition).[nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html#numpy.nonzero)(). Using nonzero directly should be preferred, as it behaves correctly for subclasses. The rest of this documentation covers only the case where all three arguments are provided.
```

server/dataset/cxg_dataset.py
bkmartinjr: The idea was just to have an explicit default in the code so we didn't have to rely on a TileDB default.  You are right that the current configurations override it to match our current deployment.  As or deployment evolves (eg, multi-region?), the config could change.

atolopko-czi (approved):
Approving, with a few minor comments!

PR #241, opened 2022-03-30
fix: Docker-compose specify platform to get around arm vs amd arch mismatches

edulop91:
Make sure we build/run x86 images everywhere.
See https://github.com/chanzuckerberg/czgenepi/pull/999/files 

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -4,6 +4,7 @@
     "default_env": "rdev",
     "app": "hosted-cellxgene",
     "default_compose_env_file": ".env.ecr",
+    "default_compose_env": ".env.ecr",
     "environments": {
         "rdev": {
             "aws_profile": "single-cell-dev",
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -3,7 +3,7 @@
     "terraform_version": "0.13.5",
     "default_env": "rdev",
     "app": "hosted-cellxgene",
-    "default_compose_env": ".env.ecr",
+    "default_compose_env_file": ".env.ecr",
     "environments": {
         "rdev": {
             "aws_profile": "single-cell-dev",
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,6 @@ services:
 
   localstack:
     image: localstack/localstack@sha256:7c6635493185d25165979995fb073fd789c72b6d8b17ef3a70b798d55576732f
-    platform: linux/amd64
     ports:
       - "4566:4566"
     environment:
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
 
   localstack:
     image: localstack/localstack@sha256:7c6635493185d25165979995fb073fd789c72b6d8b17ef3a70b798d55576732f
+    platform: linux/amd64
```
atolopko-czi, line 7: probably not necessary for `localstack`.  if running container on M1, and an arm64 platform image is available, that should be fine to use (and faster, presumably).  in general, a platform only matters for containers we're building ourselves.

docker-compose.yml
```diff
@@ -17,12 +18,13 @@ services:
...
-       corporanet:
-          aliases:
-            - localstack.corporanet.local
+      corporanet:
+        aliases:
+          - localstack.corporanet.local
 
   explorer:
     image: "${DOCKER_REPO}corpora-explorer"
+    platform: linux/amd64
```
atolopko-czi, line 27: 👍

.happy/config.json
```diff
@@ -3,7 +3,7 @@
     "terraform_version": "0.13.5",
     "default_env": "rdev",
     "app": "hosted-cellxgene",
-    "default_compose_env": ".env.ecr",
+    "default_compose_env_file": ".env.ecr",
```
atolopko-czi, line 6: Is this a happy golang related change or was it just broken? If the former, do we also need to update config_version to v2? (per [these docs](https://czi.atlassian.net/wiki/spaces/SI/pages/2429288766/Golang+Happy+CLI#Migrating-existing-repos-to-Golang-Happy-CLI))

atolopko-czi (changes requested):
a couple minor considerations

docker-compose.yml
edulop91, line 7:
```suggestion
```

.happy/config.json
edulop91: >If the former, do we also need to update config_version to v2? (per [these docs](https://czi.atlassian.net/wiki/spaces/SI/pages/2429288766/Golang+Happy+CLI#Migrating-existing-repos-to-Golang-Happy-CLI))
we will do those once we've migrated

edulop91, line 6:
```suggestion
    "default_compose_env_file": ".env.ecr",
    "default_compose_env": ".env.ecr",
```
I will keep both for now so that we don't break python

atolopko-czi (approved):

.happy/config.json
atolopko-czi: oh woops, maybe I should have been the one to accept this suggestion prior to approval? looks like it merged w/o it.

PR #244, opened 2022-04-04
Add high performance diffex routes

bkmartinjr:
#### Reviewers
**Functional:**   @atolopko-czi 
**Readability:**   @ebezzi 

---
## Changes
* builds on top of broders/diffex (PR #234)
* Add binary routes for diffex, built on the previous postings list encoding package.
* Add support for _both_ list and mask parameters in the diffex algo.  This removes an unnecessary conversion to/from masks in the new routes, saving memory & cpu.
* For sparse arrays, perform diffex tiledb queries AND mean/variance calculations incrementally, reducing memory footprint.
* Reduced the configuration value for TileDB-Py query buffer allocation size to take advantage of the incremental queries.
* Reverted the hard-coded diffex "top N" limit of 15 from the backend route.  The front-end parameter is now obeyed.
* Set the "top N" limit, in the front-end, to match the current UI (15 results)
* Upgrade TileDB to 0.13.3 (from 0.13.2)
* Improve the stand-alone run_diffexp.py benchmark


client/src/actions/index.ts
```diff
@@ -240,45 +241,44 @@ const dispatchDiffExpErrors = (
...
+      const deArgs: DiffExArguments = {
+        mode: DiffExMode.TopN,
+        params: { N: num_genes },
+        set1: set1Uint32,
+        set2: set2Uint32,
+      };
 
       const res = await fetch(
-        `${globals.API.prefix}${globals.API.version}diffexp/obs`,
+        `${globals.API.prefix}${globals.API.version}diffexp/obs2`,
```
bkmartinjr, line 274: If anyone has other ideas on a route name, this can be easily changed.  Semantics are (mostly) the same as the previous route - it is the encoding that has changed.

server/app/api/v3.py
```diff
@@ -175,6 +182,7 @@ def add_resource(resource, url):
     add_resource(ColorsAPI, "/colors")
     # Computation routes
     add_resource(DiffExpObsAPI, "/diffexp/obs")
+    add_resource(DiffExpObs2API, "/diffexp/obs2")
```
bkmartinjr, line 185: If anyone has other ideas on a route name, this can be easily changed. Semantics are (mostly) the same as the previous route - it is the encoding that has changed.

ebezzi: Maybe we could introduce versioning by route? This file is `v3` but we could add a `register_api_v4` with this route only. This is probably a bit more work so I would keep this as an optional suggestion.

server/default_config.py
```diff
@@ -101,7 +101,7 @@
       # data_locator / s3 / region_name.
       tiledb_ctx:
         sm.tile_cache_size:  8589934592  # 8GiB
-        py.init_buffer_bytes: 1073741824  # 1GiB
+        py.init_buffer_bytes: 536870912  # 512MiB
```
ebezzi, line 104: Is this the recommended configuration for the server? As far as I can remember, this configuration file lives in the single-cell-infra repo (see [here](https://github.com/chanzuckerberg/single-cell-infra/blob/main/terraform/modules/hosted-cellxgene/common/customize-prod/config.yaml)) and I am not quite sure what this python file is for, so we might want to move this to the other file.

ebezzi (approved):

server/default_config.py
bkmartinjr: This is the default used when run as a stand-alone (development) server on your own host/laptop.  Ie, from the CLI.

The deployed config is indeed elsewhere, and should be tuned for the deployment environment.  While the Python buffer size is likely in-common, the tile cache will vary by host

ebezzi: That makes sense, thanks!

server/app/api/v3.py
bkmartinjr: I will follow your lead.  The old API is going away at some point in the near future.  If that is best implemented by versioning, then it makes sense to me.  I didn't use that approach as the API is completely incompatible due to the new encoding - ie, it isn't just a revision of the old API.  It is a new API implementing the same functionality.

ebezzi: @atolopko-czi what is your preference? If we plan to remove the old route after the release, we can just switch the new one to `obs` (no conflict risk since BE+FE are deployed together) so I'd skip versioning.

server/compute/diffexp_cxg.py
```diff
@@ -27,18 +27,33 @@ def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01, arr="
...
-    :param maskB: observation selection mask for set 2
+    :param setA: observation selector (mask or list) for set 1
+    :param setB: observation selector (mask or list) for set 2
     :param top_n: number of variables to return stats for
     :param diffexp_lfc_cutoff: minimum
     :param arr: the tdb array to read
+    :param selector_lists: if True, the selectors are presumed to be masks of length n_obs; else, lists of obs indices
     absolute value returning [ varindex, logfoldchange, pval, pval_adj ] for top N genes
     :return:  for top N genes, {"positive": for top N genes, [ varindex, foldchange, pval, pval_adj ],
               "negative": for top N genes, [ varindex, foldchange, pval, pval_adj ]}
```
atolopko-czi, line 38: Not changed on this PR, but I can't quite parse the above; should it be as follows?

```suggestion
    :return:  for top N genes, {"positive": [ varindex, foldchange, pval, pval_adj ], "negative": [ varindex, foldchange, pval, pval_adj ]}
```

atolopko-czi (approved):
Minor comments only. 

We could remove quite a bit more code related to the old diffexp mechanism, if we completely remove (or replace) the old /diffexp/obs endpoint. Certainly something we could do on a future story, but if we don't do it now, might not get done.

server/compute/diffexp_cxg.py
```diff
@@ -27,18 +27,33 @@ def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01, arr="
...
     """
     matrix = adaptor.open_array(arr)
-    row_selector_A = maskA.nonzero()[0]
-    row_selector_B = maskB.nonzero()[0]
+    dtype = matrix.dtype
+    n_obs = matrix.shape[0]
+    cols = matrix.shape[1]
+    is_sparse = matrix.schema.sparse
```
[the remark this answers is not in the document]
bkmartinjr: yes, we still actively use dense.

Issue #213, opened 2022-01-20
Clean up e2eAnnotations.test.ts


atolopko-czi:
There are tests in `client/__tests__/e2e/e2eAnnotations.test.ts` that are related to _user_ annotations, which are no longer supported in hosted Explorer. These can be removed.

There are also tests, listed below, that continue to have value and these are not currently being run by any GHA job. These should be kept, and probably moved into more appropriately-named test files. And they should be run by the GHA `push_tests.yml` `smoke-tests` job.

* "geneSET crud operations and interactions"
* "GENE crud operations and interactions"
* "stacked bar graph renders"
  * Need to look into this one, but can't remember why this was implemented as a annnotations test
 * "truncate midpoint whitespace"
    * Easily added as a new category to test dataset
 * "truncate single character"
    * Likewise

_Originally posted by @seve in https://github.com/chanzuckerberg/single-cell-explorer/pull/180#r787181063_

PR #281, opened 2022-06-11
chore: clean up e2eAnnotations.test.ts #213

atarashansky:
Issue #213 

#### Reviewers
**Functional:** 
@seve 

**Readability**
@atolopko-czi 

## Changes
- removed irrelevant tests as annotations are no longer modifiable in any way
- added `smoke-test-annotations` to `push_tests.yml`


client/__tests__/e2e/e2eAnnotations.test.ts
```diff
@@ -551,65 +364,4 @@ describe.each([
 
     expect(result).toMatchSnapshot();
   });
-
-  test("truncate midpoint whitespace", async () => {
```
atarashansky, line 555: @seve Is this something that needs to be tested? I don't think labels can be renamed. Is there automatic whitespace truncation when loading annotations into the frontend? If so, I'll add the test back.

client/__tests__/e2e/e2eAnnotations.test.ts
```diff
@@ -551,65 +364,4 @@ describe.each([
...
-    const newLabelName = "123 456";
-    await renameLabel(perTestCategoryName, perTestLabelName, newLabelName);
-    const value = await waitByID(
-      `categorical-value-${perTestCategoryName}-${newLabelName}`
-    );
-    const result = await page.evaluate((elem) => elem.outerHTML, value);
-    expect(result).toMatchSnapshot();
-  });
-
-  test("truncate single character", async () => {
```
atarashansky, line 566: Same question here as for L555.

.github/workflows/push_tests.yml
```diff
@@ -103,3 +103,7 @@ jobs:
         run: |
           cd client && make smoke-test
           ./node_modules/codecov/bin/codecov --yml=../.codecov.yml --root=../ --gcov-root=../ -C -F frontend,javascript,smokeTest
+      - name: Smoke tests (with annotations feature)
```
atarashansky, line 106: I couldn't find where `e2eAnnotations.test.ts` was being run, so I added it to `push_tests.yml`.

atolopko-czi: It's be a while since I originally filed the story behind this PR, so can you remind me what UI features this is actually testing? Is it "gene sets"?  Since user annotations have been removed, that's all I can imagine it is doing. 

I'm not privy to the history behind why these two smoke test groups were split in the first place. If there's no good reason (any more), I'm wondering if it would be worth merging them together to simplify things?

atarashansky: > t's be a while since I originally filed the story behind this PR, so can you remind me what UI features this is actually testing? Is it "gene sets"? Since user annotations have been removed, that's all I can imagine it is doing.

At this point, the annotations tests are covering:
- the possible actions related to genesets (creation, coloring, editing, undo/redo, etc)
- differential expression (not sure why that's being done here but I suppose it relates to geneset creation)
- stacked bar graph renders (relates to the categorial labels so I guess it belongs here)
- check the number of cells associated with a label

> I'm not privy to the history behind why these two smoke test groups were split in the first place. If there's no good reason (any more), I'm wondering if it would be worth merging them together to simplify things?

Me neither - I think merging the tests makes sense.

atolopko-czi: Great, thanks for listing those. Those all seem like core features to test so merging all the smoke tests together still sounds right. @czi-sci-single-cell-eng Anyone disagree or have insights as to why they were split in the first place?

MDunitz: I think they were originally split prior to the repo split. Because annotations were only available if specified in the config (usually for the desktop) we wanted to make sure the general flow worked with and without them. It should be safe to merge at this point

seve: ditto, safe to merge

client/__tests__/e2e/e2eAnnotations.test.ts
seve: see https://github.com/chanzuckerberg/cellxgene/issues/1578

atarashansky:
I merged the test files, but it looks like none of the tests are working due to dependency issues with webpack in GHA. I'll create a new issue for this (@seve, if one already exists feel free to close the one I make). Until then, this PR is blocked.

atolopko-czi (commented):
Do we also need to remove this?

```
❯ ag -i e2e-annotations
client/package.json
12:    "e2e-annotations": "jest --config __tests__/e2e/e2eJestConfig.json e2e/e2eAnnotations.test.ts",
```

atolopko-czi (approved):
One additional removal noted in comment, but approving as is.

atarashansky:
It seems like these annotations tests were never being run by GHA because they simply don't work... Investigating now.

Issue #275, opened 2022-06-03
Add events to hosted explorer to track engagement


signechambers1:
We'd like to have metrics that capture whether our users get value from explorer so we can track trends in overall engaged users. For explorer, we are defining this as the number of visitors that take any of the following actions:

- Color a category, standard or author categories
- Run differential expression
- Add a gene and color by that gene
- Add a gene and plot on x/y
- Add a gene + selecting via the histogram

To track this accurately, we need to custom events for the following actions:
- Add a gene and color by that gene
- Add a gene and plot on x/y
- Add a gene + selecting via the histogram

cc @ainfeld

atarashansky:
The actions described already have events  attached to the individual "subactions":

for example:

Adding a gene using the quick search:
`EVENTS.EXPLORER_SUGGEST_MENU_ITEM_CLICKED`

Coloring by a gene:
`EVENTS.EXPLORER_COLORBY_GENE_BUTTON_CLICKED`

Clicking x for the scatterplot: 
`EVENTS.EXPLORER_PLOT_X_BUTTON_CLICKED`

Clicking y for the scatterplot:
`EVENTS.EXPLORER_PLOT_Y_BUTTON_CLICKED`

So "Add a gene and color by that gene" would appear as the `EXPLORER_SUGGEST_MENU_ITEM_CLICKED` event followed by the `EXPLORER_COLORBY_GENE_BUTTON_CLICKED` event.
Do the above sequence of events need to be displayed as a single event?

Also, in "Adding a gene and selecting via histogram"  - could you clarify what you mean by selecting via histogram?

signechambers1:
Yup, plausible makes things a little difficult, we do need those sequence of events to be displayed as a single event and yes, the order matters, they would need to add a gene first.

atarashansky:
Got it. Two more clarifying questions:

### Question 1

So if a user does:
1) Add a gene using quick search
2) Color by a gene
This would show up as something like `ADD_GENE_AND_COLORBY`

But what if a user does:
1) Add a gene using quick search
2) Something else
3) Color by a gene
Should this trigger the `ADD_GENE_AND_COLORBY` event?

### Question 2

> - Add a gene + selecting via the histogram
Do you mean cross-filtering using the brushable histogram?

signechambers1:
Thanks @atarashansky, great questions.

Question 1: Yes, i think the second scenario should also trigger the ADD_GENE_COLORBY event if possible
[...]

PR #283, opened 2022-06-14
feat: added add-gene-and-* analytics for issue 275

atarashansky:
Issue: #275 

#### Reviewers
**Functional:** 
@seve 
**Readability:** 
@seve 
---

## Changes
added analytics to track:
1. add gene and colorby: `EXPLORER_ADD_GENE_AND_COLORBY`
2. add gene and select histogram: `EXPLORER_ADD_GENE_AND_SELECT_HISTOGRAM`
3. add gene and display scatterplot: `EXPLORER_ADD_GENE_AND_DISPLAY_SCATTERPLOT`
4. select histogram: `EXPLORER_SELECT_HISTOGRAM`
5. display scatterplot: `EXPLORER_DISPLAY_SCATTERPLOT`


seve (changes requested):
Since we're editing these files, let's type where we can.  I think our _light requirement_ is function scope typing. OFC if the function is massive, just type at your own discretion.

client/src/components/brushableHistogram/index.tsx
```diff
@@ -55,6 +55,7 @@ const HEIGHT_MINI = 15 - MARGIN_MINI.TOP - MARGIN_MINI.BOTTOM;
     isColorAccessor:
       (state as any).colors.colorAccessor === field &&
       (state as any).colors.colorMode !== "color by categorical metadata",
+    userDefinedGenes: (state as any).controls.userDefinedGenes,
```
seve, line 58: Instead of adding additional redux state to the component and causing re-renders on every mutation on `userDefinedGenes`, I wonder if it would be more performant to instead do this as within a reducer? This isn't an ask, but rather an ask. Might be worth writing up a ticket

atarashansky: Consolidated in the latest version.

atarashansky:
@seve Does the new reducer (`quickGenes.ts`) look good?

PR #299, opened 2022-06-27
fix: enable deployment on rdev

roaga:
#### Reviewers
**Functional:** 
@ebezzi 

**Readability:** 
@Bento007 

---

## Changes
- modify config files to allow Explorer to be deployed on rdev.
- modify Docker-related files to take advantage of caching and trim extra files.

### DIFF
--- a/.happy/terraform/envs/staging/main.tf
+++ b/.happy/terraform/envs/staging/main.tf
@@ -17,6 +17,8 @@ module stack {
   stack_prefix                 = "/${var.stack_name}"
   batch_container_memory_limit = 28000
   memory                       = 50000
+  explorer_instance_count      = 2
+
 
   api_domain                   = "api.${local.domain}"
   web_domain                   = "${local.domain}"


hosted/Dockerfile
```diff
@@ -1,40 +1,46 @@
-FROM ubuntu:focal
+FROM ubuntu:focal as build
 
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 ARG HAPPY_ENV
 ENV HAPPY_ENV=$HAPPY_ENV
+ENV STACK_NAME="roaga-somaexplorer"
```
ebezzi, line 7: Did the infra team reach out to you about whether this variable is available at build time?

hosted/Dockerfile
```diff
@@ -1,40 +1,46 @@
...
 RUN apt-get update && \
     apt-get install -y gettext moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq npm git zip && \
-    rm -rf /var/lib/apt/lists/*
-
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
-
-RUN python3 -m pip install psycopg2-binary
+    rm -rf /var/lib/apt/lists/* && \
+    apt-get clean && \
+    update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1 && \
```
ebezzi, line 16: Wondering if this is still necessary. I believe `focal` already has python3.8 as default.

hosted/Dockerfile
```diff
@@ -1,40 +1,46 @@
...
+    python3 -m pip install psycopg2-binary
```
ebezzi, line 17: I would remove this. `psycopg2` is a driver for Postgres which isn't used by the explorer.

hosted/Dockerfile
roaga: It's not available directly. "You could add those to your task definition as environment variables, which could then be used by your docker-compose." But that's not any easier for a user of this system than just changing this file. It is a feature they are thinking about though and are looking for feedback on.

docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
       context: .
       dockerfile: hosted/Dockerfile
       cache_from:
-        - "${DOCKER_REPO}corpora-explorer:branch-main"
+        - "${DOCKER_REPO}corpora-explorer:latest"
```
ebezzi, line 31: Any reason for this?

roaga: The tag on the images is "latest," not "branch-main," so when it tried to pull from cache, it often errored that it couldn't find it. Switching to "latest" fixes that.

ebezzi: Ah sorry, this is just for caching. 👍 then

ebezzi (approved):

PR #310, opened 2022-07-06
chore: runtime config

ebezzi:
#### Reviewers
**Functional:** 
@roaga @MDunitz @atolopko-czi 

**Readability:** 

---

## Changes
- Moves configuration to runtime. Variables are now defined in the Terraform repos (as they should, since they're environment specific), added to the ECS tasks as environment variables, and interpolated inside the `config.yaml` file using a library.


server/eb/app.py
```diff
@@ -149,6 +149,7 @@ def get_csp_hashes(app, app_config):
             if config_location.exists():
                 with config_location.local_handle() as lh:
                     logging.info(f"Configuration from {config_file}")
+                    print(lh)
```
MDunitz, line 152: just curious -- why print this versus logging it?

server/eb/app.py
```diff
@@ -159,12 +160,13 @@ def get_csp_hashes(app, app_config):
...
         sys.exit(1)
 
+    # TODO: looks deprecated - remove
     dataroot = os.getenv("CXG_DATAROOT")
     if dataroot:
         logging.info("Configuration from CXG_DATAROOT")
         app_config.update_server_config(multi_dataset__dataroot=dataroot)
 
-    # overwrite configuration for the eb app
+    # TODO: this doesn't make sense
```
MDunitz, line 169: I think we only support cxg matrices in the explorer at this point?

MDunitz (approved):
lgtm

server/eb/app.py
ebezzi: I will leave it just in case.

PR #326, opened 2022-07-20
fix: added flag to avoid calculating centroid if overlay isn't displayed

atarashansky:
Calculating centroids took ~75% of every first color-by. For example, in the 4m cell dataset, coloring by a categorical for the first time took 2 seconds, 1.5 of which are the centroids. 

This PR makes it so centroids are only calculated if the overlay is displayed.

Part of this epic: #327


bkmartinjr (approved):
LGTM

seve:
Great idea!

seve:
Wait, shouldn't the result of calcCentroid been memoized?

atarashansky:
It is memo-ized but the first colorby is much slower because it's calculated by default.

PR #328, opened 2022-07-21
fix: update colorTable code so linear scan only happens once

atarashansky:
This PR removes `computePointColors` from the graph and scatterplot components. It was a redundant function that was required to translate an RGB array (an array of arrays of length 3) into a contiguous Float32Array of length #cells*3. Here, we cut out the middleman and directly calculate `colorTable.rgb` to be the Float32Array.

### DIFF
--- a/client/src/util/stateManager/colorHelpers.ts
+++ b/client/src/util/stateManager/colorHelpers.ts
@@ -23,7 +23,7 @@ import {
 
 interface Colors {
   // cell label to color mapping
-  rgb: [number, number, number][];
+  rgb: Float32Array;
   // function, mapping label index to color scale
   scale: d3.ScaleSequential<string, never> | undefined;
 }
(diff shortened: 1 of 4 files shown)


client/src/util/stateManager/colorHelpers.ts
```diff
@@ -87,9 +87,9 @@ export function createColorQuery(
 }
 
 function _defaultColors(nObs: number): Colors {
-  const defaultCellColor = parseRGB(globals.defaultCellColor);
+  // const defaultCellColor = parseRGB(globals.defaultCellColor);
   return {
-    rgb: new Array(nObs).fill(defaultCellColor),
+    rgb: new Float32Array(3 * nObs).fill(0),
```
bkmartinjr, line 92: this breaks the global defaultCellColor config. Suggest maintaining that, OR if you find the global not useful, cleaning it up (the comment on line 90, and the line in globals.ts)

bkmartinjr (commented):
Approving, but with a couple of asks pre-merge:
* add a PR description for posterity
* clean up the breakage around the global default color

caveat:  I did a code read, but did not verify or test that there are not other dependencies affected by the change (ie, other code that uses `rgb`).

bkmartinjr (approved):

client/src/util/stateManager/colorHelpers.ts
atarashansky: Ah shoot, I forgot to fix this.

PR #330, opened 2022-07-21
Atar/better fbs

atarashansky:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify


server/common/fbs/matrix.py
```diff
@@ -52,21 +54,22 @@ def serialize_typed_array(builder, source_array, encoding_info):
 
     arr = source_array
     (array_type, as_type) = encoding_info(source_array)
-
     if isinstance(arr, pd.Index):
         arr = arr.to_series()
 
     # convert to a simple ndarray
-    if as_type == "json":
```
atarashansky, line 60: I need to add this back to catch edge cases (not Categorical and not a typed array, this is probably for columns like cell barcode IDs)

atolopko-czi (changes requested):
I think it would be better to upgrade explorer to use Go happy, since `happy push` will then build and push images, as is done in data portal [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.github/workflows/push-rdev.yml#L45). @ebezzi Any reason we can't upgrade to Go happy in explorer?

atarashansky:
> I think it would be better to upgrade explorer to use Go happy, since `happy push` will then build and push images, as is done in data portal [here](https://github.com/chanzuckerberg/single-cell-data-portal/blob/main/.github/workflows/push-rdev.yml#L45). @ebezzi Any reason we can't upgrade to Go happy in explorer?

I updated the workflow to use the happy go client.

PR #337, opened 2022-07-22
Updating the flatbuffer scheme with dictionary-encoded and sparse data types

atarashansky:
The Flatbuffer schema has been updated with new data structures for the efficient transfer of categorical data over the wire. With the addition of more complex data types that require specialized coding routines, the FB encoders and decoders have been encapsulated into different classes and/or functions for the three main data types: categorical, dense numeric, and polymorphic.

Dictionary-encoded arrays are used to efficiently store categorical data types, typically strings. The underlying array structure is composed of integer “codes”, where each code corresponds to one of the unique strings or “levels”. They are attached to small JSON blobs that map the codes to the levels.

### Backend changes:
 - Updated the encoder/decoders logic in `matrix.py` and `fbs_coders.py`.
 - When loading obs data from TileDB, if the schema hint says it is categorical, it is converted into pandas Categorical. This is a temporary measure as in the future, the obs data will hopefully be dictionary encoded in the CXG itself.

### Frontend changes:
 - Updated the encoder/decoders logic in `matrix.ts`.
 - `dataframe.ts` has been updated with a new column type (`DataframeDictEncodedColumn`) that is a subtype of `DataframeColumn`.
 - `DataframeDictEncodedColumn`s have properties specific to dictionary encoded arrays, such as `codeMapping` (codes-to-values), `invCodeMapping` (values-to-codes).
 - `<DataframeDictEncodedColumn>.asArray()` returns the underlying array, which yields codes, not values. As a result, anywhere the client needs to display annotation labels, we are now checking to see if it is dictionary encoded and then displaying the mapped values instead of the codes using the `codeMapping` object. Functions like `at` and `iat` that are specifically designated in docstrings to return values have been updated to return the appropriate values for dictionary encoded columns.

### Note to reviewers
You can ignore the changes in `server/common/fbs/NetEncoding/` and `client/src/util/stateManager/net-encoding/` as these are automatically generated by the flatbuffer compiler.


client/src/util/dataframe/dataframe.ts
```diff
@@ -128,6 +138,7 @@ class Dataframe {
     columnarData: DataframeValueArray[],
     rowIndex?: LabelIndex | null,
     colIndex?: LabelIndex | null,
+    columnDicts?: { [key: string]: { [key: number]: string } } | null,
```
atarashansky, line 141: NTS: type this with DictEncoder

client/src/util/dataframe/dataframe.ts
```diff
@@ -215,6 +229,7 @@ class Dataframe {
   /** @internal */
   private static __compileColumn(
     column: DataframeValueArray,
+    columnDict: { [key: number]: string },
```
atarashansky, line 232: NTS: create a type for this

fbs/matrix.fbs
```diff
@@ -22,39 +22,70 @@
...
-  Float64Array,
-  JSONEncodedArray
+table CatInt8FBArray {
+  data: [int8];
+  codes: [uint8];
+}
+
+table CatInt16FBArray {
+  data: [int16];
+  codes: [uint8];
```
bkmartinjr, line 54: please document (comment) the assumed encoding of the code dictionary.  Eg,
* JSON object, UTF-8, encoded as a byte array
* JSON object schema is ???? (array of ..., dict of ..., ...?)

fbs/matrix.fbs
```diff
@@ -22,39 +22,70 @@
...
+}
+
+table CatInt32FBArray {
+  data: [int32];
+  codes: [uint8];
+}
+
+table SparseFloat32FBArray {
```
bkmartinjr, line 62: please add comment describing data structure.  I can't quite discern if this is a COO matrix, or something else.

fbs/matrix.fbs
```diff
@@ -22,39 +22,78 @@
...
-  Float32Array,
-  Int32Array,
-  Uint32Array,
-  Float64Array,
-  JSONEncodedArray
+table CatInt8FBArray {  
+  data: [int8];
+  // "codes" is a UTF-8/JSON dictionary encoded as a byte array,
+  // Stores the mapping between codes and values
+  // JSON object schema is {[code: integer]: value: string}
```
bkmartinjr, line 51: is `value` in the code table _always_ a string, or do you also use this type for other types (boolean, int, etc)?  Pandas Categoricals support all of the above which is why I ask.

If it is truly used _only_ for string categoricals, then you would likely benefit from encoding the code table as a flatbuffer array of struct, where the struct is { int, string}.  If it is actually polymorphic, then JSON makes sense.

In the past, we have seen categorical arrays that had int and bool category types.  But the overwhelming majority is string.  TL;DR - there may be more optimization available here if you want it.

atarashansky: The intention was to allow polymorphic types that would get converted to strings in the fronted - typing the value as string here was an inaccuracy.

fbs/matrix.fbs
```diff
@@ -7,8 +7,8 @@
   organized in a columnar layout.   Each column is homomorphic, and
```
bkmartinjr, line 7: comment is obsolete - please update

client/package.json
```diff
@@ -25,7 +25,8 @@
     "extends": "./configuration/eslint/eslint.js"
   },
   "eslintIgnore": [
-    "src/util/stateManager/matrix_generated.js"
+    "src/util/stateManager/matrix_generated.js",
```
bkmartinjr, line 28: this file is no longer machine generated, so it should not be ignored by the linter

server/common/fbs/matrix.py
```diff
@@ -1,5 +1,5 @@
 import json
-
+from json import JSONDecodeError
```
bkmartinjr, line 2: this import is not used, so it is not needed.

suggest linting the code

fbs/matrix.fbs
```diff
@@ -22,39 +22,78 @@
 
 namespace NetEncoding;
 
-table Float32Array {
+table Float32FBArray {
```
bkmartinjr, line 25: the addition of `FB` to clarify that this is a FB type is good -- adds clarity, but I suggest you do it uniformly to all types defined in this file.

it also might be clearer if it is a prefix, eg, FBFloat32Array, FBColumn, etc. (I don't feel strongly about this)

server/common/fbs/matrix.py
```diff
@@ -8,14 +8,18 @@
...
-
+import server.common.fbs.NetEncoding.TypedFBArray as TypedFBArray
+import server.common.fbs.NetEncoding.Uint32FBArray as Uint32FBArray
+import server.common.fbs.NetEncoding.CatInt16FBArray as CatInt16FBArray
+import server.common.fbs.NetEncoding.CatInt32FBArray as CatInt32FBArray
+import server.common.fbs.NetEncoding.CatInt8FBArray as CatInt8FBArray
+import server.common.fbs.NetEncoding.SparseFloat32FBArray as SparseFloat32FBArray
+import server.common.fbs.NetEncoding.SparseFloat64FBArray as SparseFloat64FBArray
 
 # Serialization helper
```
bkmartinjr, line 24: another flake8 error here - please lint before commit.

server/common/fbs/matrix.py
```diff
@@ -46,74 +50,91 @@ def serialize_matrix(builder, n_rows, n_cols, columns, col_idx):
...
-        np.dtype(np.int8).str: (TypedArray.TypedArray.Int32Array, np.int32),
-        np.dtype(np.int16).str: (TypedArray.TypedArray.Int32Array, np.int32),
-        np.dtype(np.int32).str: (TypedArray.TypedArray.Int32Array, np.int32),
-        np.dtype(np.int64).str: (TypedArray.TypedArray.Int32Array, np.int32),
-        np.dtype(np.uint8).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
-        np.dtype(np.uint16).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
-        np.dtype(np.uint32).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
-        np.dtype(np.uint64).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
+        # array protocol string:  ( array_type, as_type1, as_type2 )
+        # as_type2 is used for flatbuffers with more than one slot
```
bkmartinjr, line 105: this comment needs additional clarification.  I also suggest improved naming for the fields (maybe used a NamedTuple to be self-documenting?)

server/common/fbs/matrix.py
```diff
@@ -46,74 +50,91 @@ def serialize_matrix(builder, n_rows, n_cols, columns, col_idx):
...
     arr = source_array
-    (array_type, as_type) = encoding_info(source_array)
-
+    (array_type, as_type1, as_type2) = encoding_info(source_array)
     if isinstance(arr, pd.Index):
         arr = arr.to_series()
 
-    # convert to a simple ndarray
-    if as_type == "json":
+    types = [as_type1, as_type2, "int"]
```
bkmartinjr, line 62: this hard-wires the assumption that the third slot is _always_ an int.  Ie, that is a very hidden shared dependency between the matrix.fbs file and this code.  Seem incredibly error prone - any change to the FBS file will break this.

I suggest you either:
1. add the third slot to the encoding table
2. comment the heck out of this in _both_ files

I prefer the first, but that is just an opinion. Even the encoding table is a shared assumption between the two files (just trying to put all of the assumptions in one place)

server/common/fbs/matrix.py
```diff
@@ -46,74 +50,91 @@ def serialize_matrix(builder, n_rows, n_cols, columns, col_idx):
...
+    if as_type1 == "json":
         as_json = arr.to_json(orient="records")
         arr = np.array(bytearray(as_json, "utf-8"))
+        fields = [arr]
     else:
         if sparse.issparse(arr):
-            arr = arr.toarray()
+            row = arr.row
+            data = arr.data
+            size = np.array([max(arr.shape)], dtype="int")
```
bkmartinjr, line 72: the FB schema defines this field as an int32, but here you are creating an int64. Quite possibly a bug, or you are later doing an unnecessary cast to int32.  Either way, shouldn't this read `dtype="int32"`?

fbs/matrix.fbs
```diff
@@ -22,39 +22,78 @@
...
+
+// a sparse-encoding of float32 arrays in
+// COO sparse matrix format
+table SparseFloat32FBArray {
+  // nonzero data
+  data: [float32];
+  // nonzero locations of the data in the array
+  rows: [int32];
+  // size of the array
+  size: [int32];
```
bkmartinjr, line 73: why is this slot an array rather than just an int32?

fbs/matrix.fbs
```diff
@@ -22,39 +22,78 @@
...
+  codes: [uint8];
+}
+
+// a sparse-encoding of float32 arrays in
+// COO sparse matrix format
+table SparseFloat32FBArray {
+  // nonzero data
+  data: [float32];
+  // nonzero locations of the data in the array
+  rows: [int32];
```
bkmartinjr, line 71: Should `rows` be uint32 - ie, are negative row indices used?

server/common/fbs/matrix.py
atarashansky: Great suggestion. I'm on it.

server/common/fbs/matrix.py
atarashansky: Thanks for the catch - will fix.

server/dataset/cxg_dataset.py
```diff
@@ -271,10 +272,15 @@ def get_X_array(self, obs_mask=None, var_mask=None):
 
             nrows, obsindices = self.__remap_indices(X.shape[0], obs_mask, data.get("coords", data)["obs"])
             ncols, varindices = self.__remap_indices(X.shape[1], var_mask, data.get("coords", data)["var"])
-            densedata = np.zeros((nrows, ncols), dtype=self.get_X_array_dtype())
-            densedata[obsindices, varindices] = data[""]
-
-            return densedata
+            if obsindices.size / nrows > 0.5:
```
bkmartinjr, line 275: I think this is the wrong place to put this logic.  The decision being made is how to encode in a flatbuffer, which belongs inside the flatbuffer encoder (eg, in matrix.py), not in this code which is used for many purposes (eg, internal computation like diffex, etc).

Put another way, the matrix encoder should introspect the data and decide on the best FB encoding to use.

server/dataset/cxg_dataset.py
```diff
@@ -403,7 +409,14 @@ def annotation_to_fbs_matrix(self, axis, fields=None):
                 else:
                     data = A.query(attrs=fields)[:]
 
+                categorical_dtypes = []
```
bkmartinjr, line 412: same question as above - why is this not being done in the FBS encoder?  It is directly related to the FBS optimizations, not to the data model or CXG.

You are using the schema as a hint for what might have originally been a Pandas Categorical, but that hint could be passed from this layer to the underlying FBS encoder.

Perhaps even better the FBS encoder could decide on its own whether the data is better as a categorical (eg, by counting unique values in the column). The schema (at this level) is simply a hint that indicates that the original H5AD used a categorical, and that is of mixed value at best (ie, it is semi-random whether or not it is most efficiently encoded as a categorical).

server/common/utils/type_conversion_utils.py
```diff
@@ -117,23 +122,19 @@ def _get_type_info(array: Union[np.ndarray, pd.Series, pd.Index]) -> Tuple[np.dt
...
-            # NA/NaN (missing or undefined) categories.
-            if dtype.categories.dtype.kind in ["f", "i", "u"]:
-                return (
-                    _get_type_info(array.to_numpy())[0],
-                    {"type": "categorical"},
-                )
-            else:
-                return (np.dtype(str), {"type": "categorical", "categories": dtype.categories.to_list()})
+            return (
+                "cat" + str(array.cat.codes.dtype.itemsize * 8),
```
bkmartinjr, line 135: This violates the signature for the function as "cat8"  is not a numpy.dtype. Feels hackish and breaks the abstraction.

One possible solution is to return a real dtype, ie, follow the Pandas lead and return a  pd.CategoricalDtype (which contains the type of the underlying category data).

bkmartinjr (changes requested):
I have made it through the server-side (python) changes.  Will come back for a second pass on the front-end.

Suggest several improvements:
* the encoding heuristics for picking the FB array type should be solely within the matrix encoding layer (matrix.py) not spread across multiple files/layers. Look for a cleaner abstraction that doesn't require code in cxg_adaptor.py, matrix.py, etc.
* the serialize code in matrix.py is getting complex and hard to understand. There are necessary dependencies between this file and matrix.fbs, but can this be made clearer / more modular?
* the change to the behavior of _get_type_info could be improved / cleaner.  Pandas already has a dtype for categorical, so why not use it?
* code hygiene - improvement to select comments, linting, etc. Nothing major, but need to make this easier to understand for the next person who fixes a bug in it.  For example, the selection of FB array type is based on a moderately complex heuristic which needs solid comments in matrix.py.

client/src/util/stateManager/matrix.ts
```diff
@@ -12,30 +12,68 @@ import {
...
   // @ts-expect-error --- FIXME: Element implicitly has an 'any' type.
-  const TypeClass = NetEncoding[NetEncoding.TypedArray[uType]];
+  const TypeClass = NetEncoding[NetEncoding.TypedFBArray[uType]];
   // Create a TypedArray that references the underlying buffer
-  let arr = uValF(new TypeClass()).dataArray();
-  if (uType === NetEncoding.TypedArray.JSONEncodedArray) {
+  const x = uValF(new TypeClass());
+  let arr = x.dataArray();
+  let dict;
+  let flag = true;
```
bkmartinjr, line 41: please use descriptive variable names.  Eg  `release_fb_buffer` or something self-documenting.

bkmartinjr: same for `dict`, etc.  Make variable names descriptive

client/src/util/stateManager/matrix.ts
```diff
@@ -12,30 +12,68 @@ import {
...
 
 /**
  * Matrix flatbuffer decoding support. See fbs/matrix.fbs
  */
 
 /**
- * Decode NetEncoding.TypedArray
+ * Decode NetEncoding.TypedFBArray
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
```
bkmartinjr, line 28: My understanding is that the team agreement is that these eslint disables will be fixed whenever major changes are made to a file.  I definitely think that they are hiding type errors, so it is worth cleaning them up IMHO.

client/src/util/stateManager/matrix.ts
```diff
@@ -12,30 +12,68 @@ import {
...
-  const TypeClass = NetEncoding[NetEncoding.TypedArray[uType]];
+  const TypeClass = NetEncoding[NetEncoding.TypedFBArray[uType]];
   // Create a TypedArray that references the underlying buffer
-  let arr = uValF(new TypeClass()).dataArray();
-  if (uType === NetEncoding.TypedArray.JSONEncodedArray) {
+  const x = uValF(new TypeClass());
+  let arr = x.dataArray();
+  let dict;
+  let flag = true;
+  if (uType === NetEncoding.TypedFBArray.JSONEncodedFBArray) {
```
bkmartinjr, line 42: this whole section (if/else/else...) would be clearer if it was refactored into a set of stand-alone functions which decode a single type, eg,  `decodeJSONTypedArray()`, `decodeSparseTypedArray()`, etc.

client/src/util/stateManager/matrix.ts
```diff
@@ -12,30 +12,68 @@ import {
...
+    }
+    flag = false;
+  } else if (
+    uType === NetEncoding.TypedFBArray.SparseFloat32FBArray ||
+    uType === NetEncoding.TypedFBArray.SparseFloat64FBArray
+  ) {
+    const rows = x.rowsArray();
+    const size = x.sizeArray()[0];
+    const data = new arr.constructor(size);
+    for (const [i, row] of rows.entries()) {
```
bkmartinjr, line 69: is there a reason you use for-of instead of a simple for loop?  My experience is that iterators have traditionally been slower, and this is a spot you _really_ care about perf as it is a giant memory copy.

Ie,
```
for (const i = 0, nrows = rows.length; i < nrows; i += 1) {
    data[rows[i]] = arr[i];
}
```

If you do have a preference for for-of, I would do some benchmarking to make sure that it isn't (still) slower.

client/src/util/stateManager/matrix.ts
```diff
@@ -101,6 +139,9 @@ function encodeTypedArray(builder: any, uType: any, uData: any) {
 /**
  * Encode the dataframe as an FBS Matrix
  */
+
```
bkmartinjr, line 142: reminder :-)

client/src/util/stateManager/matrix.ts
```diff
@@ -242,17 +283,24 @@ export function matrixFBSToDataframe(
...
   const colIdx = fbs
     .map((b) => (Array.isArray(b.colIdx) ? b.colIdx : Array.from(b.colIdx)))
     .flat();
   const nCols = columns.length;
-
   const df = new Dataframe([nRows, nCols], columns, null, new KeyIndex(colIdx));
   return df;
 }
+
+export const isCatTypedArray = (c: unknown): boolean =>
```
bkmartinjr, line 302: wrong place for this code - it is general purpose and not specific to FB serialization.  Doesn't it belong in arraytypes.ts or something like that?

client/src/util/dataframe/types.ts
```diff
@@ -66,6 +70,11 @@ export interface DataframeColumn extends DataframeColumnGetter {
    */
   isContinuous: boolean;
 
+  /**
+   * Boolean indicating if the underlying data is dictionary-encoded.
+   */
+  isCategorical: boolean;
```
bkmartinjr, line 76: this is very, very confusing as it is a different use of the word `categorical` than the other interfaces.  Eg, `histogramCategorical` does not require a dictionary-encoded column (it works on any array/column type).

I would come up with a new term, and not overload the concept - which will lead to much confusion.

(ie, clean abstractions!)

bkmartinjr: Another thought - I think you may want to create multiple specialized sub-interfaces for DataframeColumn, so it can be used as a "type", eg,
* CategoricalDataframeColumn
* NumericDataframeColumn
* etc...
thoughts?

atarashansky: I think that's a great idea. It'd let us explicitly define attributes specific to categorical vs numerical columns.

server/common/utils/type_conversion_utils.py
atarashansky: I don't think pd.CategoricalDtype contains the type of the code array.

bkmartinjr: Try:
* for codes:    series.cat.codes.dtype
* for the lables/categories:   series.cat.categories.dtype

server/dataset/cxg_dataset.py
atarashansky: One concern I have is that returning a dense array which then needs to be converted into a sparse array would add >100 milliseconds to the request for large (>4m) datasets. How about we add an argument to `get_X_array` called `return_dense` which would return dense data regardless of density for functions that need it, like diffex.

atarashansky: Or we could do the opposite and add an `allow_sparse` argument and `density=0.5` argument to determine if `get_X_array` is allowed to return a sparse data type based on the specified density cutoff. That way, default behavior is how `get_X_array` behaved previously.

bkmartinjr: Perhaps the better trade is time of network transfer vs time to encode/decode.  It needs to be a net win across both.

One concrete suggestion:  only use the complicated encoding if:
* the array is large
* and it has low density
If both are true, then the smaller size is well worth the CPU time.  If the array is already small, the complicated encoding may not be worth it.

You will need to do some experiments to find the right threshold.

server/dataset/cxg_dataset.py
atarashansky: This is a stopgap before remaking the CXGs such that categoricals are explicitly defined in the schema and saved as integer arrays with the code-to-value dictionary. I think the determination of whether an array is / should be categorical is best done during the CXG creation process. This way, we also avoid needing to run the costly `unique` operation each time an `obs` column is fetched.

client/src/util/stateManager/matrix.ts
atarashansky: Ran into some annoying circular dependencies...

client/src/util/dataframe/types.ts
atarashansky: I think I need a primer on how types work to resolve this properly...

client/src/util/stateManager/matrix.ts
atarashansky: Decoder fixed, but encoder is giving me problems. Will keep trying.

atarashansky:
Latest changes

 - All uses of the word "Categorical" to mean dictionary encoded have been replaced with "DictionaryEncoded"
 - `DataframeColumn` is the union of `DataframeNumericColumn` (previously `DataframeColumn`) and `DataframeDictEncodedColumn` (a subclass of the former).
 - The encoding/decoding in `matrix.ts` has been organized into separate functions for each type of array (dictionary-encoded, sparse, numeric).
 - `matrix.ts` decoders are typed, now. The encoders are still a work in progress.
 - The encoding/decoding in `matrix.py` has been abstracted into `server/common/fbs/fbs_coders.py`. Each type of array has its own coding class with encoder and decoder functions.
 - Changes to `_get_type_info` have been reverted as they are no longer necessary.

client/__tests__/util/stateManager/sampleResponses.ts
```diff
@@ -4,9 +4,11 @@ import isNumber from "lodash.isnumber";
 import zip from "lodash.zip";
 import _ from "lodash";
 import { flatbuffers } from "flatbuffers";
-import { NetEncoding } from "../../../src/util/stateManager/matrix_generated";
+import { NetEncoding } from "../../../src/util/stateManager/fbs_data_types";
 import { RawSchema } from "../../../src/common/types/schema";
 
+// TODO (alec): PLEASE UPDATE THIS
```
MDunitz, line 10: remove? or do

atarashansky: oops, do. thanks!

scripts/cxg_inspect.py
```diff
@@ -21,18 +21,20 @@
...
-@click.option('--urls-file', '-f', type=str,
-              help='File containing a list of CXG file URLs, separated by newlines. '
-                   'Output will conform to jsonl format.')
-@click.option('--limit', '-l', type=int)
+@click.option("--url", "-u", type=str, help="CXG file URL. E.g. `file:/my_dataset.cxg`, or `s3://fba8204.cxg`.")
+@click.option(
+    "--urls-file",
+    "-f",
+    type=str,
+    help="File containing a list of CXG file URLs, separated by newlines. " "Output will conform to jsonl format.",
```
bkmartinjr, line 29: If a single line, I would concatenate as a single string.

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+import server.common.fbs.NetEncoding.DictEncoded32FBArray as DictEncoded32FBArray
+import server.common.fbs.NetEncoding.DictEncoded8FBArray as DictEncoded8FBArray
+import server.common.fbs.NetEncoding.SparseFloat32FBArray as SparseFloat32FBArray
+import server.common.fbs.NetEncoding.SparseFloat64FBArray as SparseFloat64FBArray
+
+
+def _get_array_class(array):
+    # return the generic type of array - category, sparse, or dense
+    # this determines which coder should be used.
+    if isinstance(array, pd.Series) and array.dtype.name == "category":
```
bkmartinjr, line 23: Are you specifically excluding the more exotic types such as Panda CategoricalIndex?

You might consider using the `pandas.api.types` helpers that make the intent more obvious, eg, [pandas.api.types.is_categorical_dtype](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_categorical_dtype.html), or pandas.api.types.is_categorical (depending on your intent).

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def serialize_typed_array(builder, source_array):
+    if isinstance(source_array, pd.Index):
+        source_array = source_array.to_series()
+
+    array_class = _get_array_class(source_array)
+
+    # depending on the array class, compute the encoding data type from
+    # the corresponding attribute of the source array. e.g. for pandas Categoricals,
```
bkmartinjr, line 38: nit:  s/Categoricals/Categorical/

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+    # the encoding data type is computed from the codes array.
+    if array_class == "category":
```
bkmartinjr, line 40: You might consider putting this logic into _get_array_class() and having that function return a tuple of `(_array_type_, _dtype_)`.  It seems to essentially use the same if/else logic.

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+class CategoricalCoder:
+    n_slots = 2
+
+    def encode_array(self, array, builder, dtype):
+        if isinstance(array, pd.Series) and array.dtype.name == "category":
+            # create the code-to-value dictionary and encode in utf-8 as a byte array
+            dictionary = np.array(bytearray(json.dumps(dict(enumerate(array.cat.categories))), "utf-8"))
+            codes = array.cat.codes.values
+
+            # this type safeguard is likely unecessary but can't hurt to include
```
bkmartinjr, line 148: spelling error

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+            if np.dtype(codes.dtype).str != dtype:
```
bkmartinjr, line 149: I'm not opposed to this, but it may hide real errors in the encoder table.  Suggest you add an assertion that the number of codes is less than max number that can be held in this code dtype.

Eg, something like:  `assert len(codes) <= np.iinfo(dtype).max + 1`

server/common/fbs/matrix.py
```diff
@@ -43,81 +38,6 @@ def serialize_matrix(builder, n_rows, n_cols, columns, col_idx):
...
-        np.dtype(np.int64).str: (TypedArray.TypedArray.Int32Array, np.int32),
-        np.dtype(np.uint32).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
-        np.dtype(np.uint64).str: (TypedArray.TypedArray.Uint32Array, np.uint32),
-    }
-    index_encoding_default = (TypedArray.TypedArray.JSONEncodedArray, "json")
-
-    return index_encoding_type_map.get(arr.dtype.str, index_encoding_default)
-
-
 def guess_at_mem_needed(matrix):
```
bkmartinjr, line 41: It would be easy to finish the implementation of this with [pandas.DataFrame.memory_usage](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.memory_usage.html), and would give you a modest perf improvement.

server/dataset/cxg_dataset.py
```diff
@@ -254,7 +255,8 @@ def __remap_indices(self, coord_range, coord_mask, coord_data):
         coordindices = mapindex[coord_data]
         return ncoord, coordindices
 
-    def get_X_array(self, obs_mask=None, var_mask=None):
+    def get_X_array(self, obs_mask=None, var_mask=None, allow_sparse_return=False):
+        # allow_sparse_return = True for /data/var requests to forego densification
```
bkmartinjr, line 259: Nit:  ideally, use the standard docstrings format please.

server/tests/unit/common/fbs/test_matrix.py
```diff
@@ -50,7 +50,7 @@ def test_encode_DataFrame(self):
                 "d": pd.Series(["x", "y", "z", "x", "y", "z", "a", "x", "y", "z"], dtype="category"),
```
bkmartinjr, line 50: This file needs to be expanded to cover all of the new FBS encoding types.  I see a bunch of missing cases (suggest looking at code coverage and ensuring you hit 100% on the matrix.py and fbscoders.py files):
* sparse float64
* sparse:  CSR & CSC
* Categoricals with code width 8, 16 and 32
etc.

server/dataset/dataset.py
```diff
@@ -72,7 +72,7 @@ def get_embedding_array(self, ename, dims=2):
         pass
 
     @abstractmethod
-    def get_X_array(self, obs_mask=None, var_mask=None):
+    def get_X_array(self, obs_mask=None, var_mask=None, allow_sparse_return=False):
```
bkmartinjr, line 75: There does not seem to be a test case for this new feature (allow_sparse_return). Please add (and check codecov to make sure you hit all the corner cases)

server/tests/unit/data_common/fbs/test_matrix.py
```diff
@@ -46,7 +46,7 @@ def test_encode_DataFrame(self):
                 "d": pd.Series(["x", "y", "z", "x", "y", "z", "a", "x", "y", "z"], dtype="category"),
```
bkmartinjr, line 46: See comment on common/fbs/test_matrix.py about needing to test the new features/encodings.

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,225 @@
...
+            np.dtype(np.int64).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Int32FBArray),
+            np.dtype(np.uint8).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
+            np.dtype(np.uint16).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
+            np.dtype(np.uint32).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
+            np.dtype(np.uint64).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
+        },
+        "sparse": {
+            np.dtype(np.float64).str: (SparseNumericCoder, TypedFBArray.TypedFBArray.SparseFloat64FBArray),
+            np.dtype(np.float32).str: (SparseNumericCoder, TypedFBArray.TypedFBArray.SparseFloat32FBArray),
+            # below is to handle cases where integer sparse arrays are encoded - densify in this case.
```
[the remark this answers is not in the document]
atarashansky: > because the matrix FBS encoding doesn't support sparse int arrays, because we have not yet found a use for them? 

Both. Expressions are the only thing we are sparse coding and they're floats. I'll clarify with a comment! Alternatively, I could add support for sparse int arrays, which may be useful if we ever wish to accommodate integer-typed raw counts, but that could also happen in a future PR. I would go with a clarifying comment for now.

scripts/cxg_inspect.py
```diff
@@ -21,18 +21,20 @@
```
MDunitz, line 22: Are all of the changes in this file linting related? Could you make a separate pr with the linting changes that are unrelated to the changes to fbs?

client/src/common/types/arraytypes.ts
```diff
@@ -2,6 +2,10 @@
  * Utility type and interface definitions.
  */
 
+import { DictEncoded16Array } from "../../util/stateManager/dict-encoded16_array";
+import { DictEncoded8Array } from "../../util/stateManager/dict-encoded8_array";
+import { DictEncoded32Array } from "../../util/stateManager/dict-encoded32_array";
+
 /**
  * TypedArrays that can be assigned to a number.
```
bkmartinjr, line 10: weird comment - I think it should say "...that can be assigned a number", or perhaps "which have numeric values"?

client/src/util/dataframe/dict-encoded.ts
```diff
@@ -0,0 +1,23 @@
+/*
+This module contains helper functions to map an array of categorical codes to values
+*/
+import { DictEncodedArray } from "../../common/types/arraytypes";
+import { DataframeValueArray } from "./types";
+
+export function mapCodesToValues(
+  column: DictEncodedArray,
+  __id: string
```
bkmartinjr, line 9: lint...  Unused parameter.

client/src/util/dataframe/dict-encoded.ts
```diff
@@ -0,0 +1,23 @@
...
+): DataframeValueArray {
+  const result: DataframeValueArray = new Array(column.length);
+  for (const [i, _] of column.entries()) {
+    result[i] = column.vat(i);
+  }
+  return result;
+}
+
+export function hashMapCodesToValues(
```
bkmartinjr, line 18: this function appears to be only used as a memoization key in dataframe.ts.  It can potentially create a huge string which seems inefficient.

Any reason you can't just stash a unique (small) ID inside the VAT array (like we already do for Dataframe objects), and then use that as the memoization key?

For example, each of the `DictEncoded*Array` classes could have a class variable (number) that is used to seed a globally unique ID which is set in the `codeMapping` object (inside the `setCodeMapping` method).

client/src/util/stateManager/dict-encoded8_array.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class DictEncoded8Array extends Int8Array {
+  codeMapping: CodeMapping;
+
+  constructor(props: ArrayBufferLike) {
+    super(props);
+    this.codeMapping = {};
+  }
+
+  setCodeMapping = (codeMap: CodeMapping) => {
```
bkmartinjr, line 18: lint nit - no return type.

client/src/util/stateManager/dict-encoded8_array.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * This module exports a subclass of Int8Array for
+ * categorical data types. The subclass contains the
+ * `codeMapping` dictionary which maps codes to values,
+ * and a helper function `vat` that returns the value at
+ * a particular index.
+ */
+import { CodeMapping } from "./code_mapping_interfaces";
+
+export class DictEncoded8Array extends Int8Array {
```
bkmartinjr, line 10: I'm really struggling with this construct from the standpoint of type system abstractions -- this class is logically an array of categories, which means it is _not_ a specialization (subclass) of an Int8Array.  Put another way, I should be able to treat this object like it is just another Int8Array, which clearly is not true in this case.

Given that the Dataframe.__compileColumn flattens the object into an Array (ie. col().values), why bother defining this type as a sub-class of Int8Array at all?

Put another way, why not just define the class as a simple object with two properties:  `codes: Int8Array` and `codeMapping: CodeMapping`?

client/src/annoMatrix/normalize.ts
```diff
@@ -140,9 +140,21 @@ export function normalizeCategorical(
   const allCategories = new Set<Category>(
     // TODO #35: Use type guards instead of casting
 
-    colDataSummary.categories.concat(
-      (colSchema as CategoricalAnnotationColumnSchema).categories ?? []
-    )
+    /** (#337) - Note for below:
+     * 1) colDataSummary.categories now contains codes
+     * 2) schema categories are values
+     * 3) user-defined/manipualted categories no longer allowed
```
bkmartinjr, line 146: spelling

client/src/annoMatrix/normalize.ts
```diff
@@ -140,9 +140,21 @@ export function normalizeCategorical(
...
+     * Consequently, the frontend state and the schema
+     * are guaranteed to have the same labels in each
+     * category, so the concatenation can be safely removed for now.
+     * Later, if user annotations are reintroduced, this will need
+     * to be uncommented. One solution is to:
+     * 1) convert schema labels that are present in the current column into new codes
+     * 2) and add these codes to the code-to-value dictionary in the categorical column
+     **/
+    colDataSummary.categories // .concat(
```
bkmartinjr, line 155: if this code is not needed, remove it.  Leaving the dangling expression is "lint"

client/src/components/categorical/value/index.tsx
```diff
@@ -69,21 +71,27 @@ type CategoryValueProps = PureCategoryValueProps & {
...
   const isDilated =
     pointDilation.metadataField === metadataField &&
     pointDilation.categoryField === _currentLabelAsString(ownProps);
 
   const category = categoricalSelection[metadataField];
   const label = categorySummary.categoryValues[categoryIndex];
-  const isSelected = category.get(label) ?? true;
+  const col = categoryData.icol(0);
+  const labelName = isDataframeDictEncodedColumn(col)
+    ? col.codeMapping[parseInt(label, 10)]
```
bkmartinjr, line 84: Reading dataframe.ts, it looks like any dict-encoded array will already have a fully flattened `values` Array containing this information.    Why do the lookup via `codeMapping` when you could use `values`?   Or perhaps, why create `values` and not use it?

atarashansky: I am not sure why I created `values` even though I had no intention to use it. I think I was still uncertain at that point about what I wanted the `DataframeDictEncodedColumn` to look like. I think we would prefer to use int typed arrays everywhere in the frontend and map the integers to string values. I will remove `values`.

client/src/util/stateManager/dict-encoded8_array.ts
atarashansky: I initially decided to subclass `Int*Array`s, so that the `.col().asArray()` would behave like an array but also contain the `codeMaping` values. This was before I created a new Dataframe column type and added the codeMapping object to the column itself, so that particular rationale may no longer apply.

The other reason I wanted to sublcass `Int*Array` was to make it easier to interface with the Dataframe constructor, as it seemed easier to pass in some type of array vs passing in an object of `codes` and `codeMapping`. In particular, anywhere that dataframe uses `__columns` internally, I would need to add verbose logic to handle that object. In the current implementation, the only place in dataframe we need to modify outside `__compileColumn` is `_subset`.

atarashansky: For example, in `at`, we return:
```typescript
return this.__columns[coff][roff];
```
This works out of the box with a subclass, but will not work with an object. 

Also, a particularly challenging spot for me when trying the object interface was in `mapColumns`: the logic for dealing with a dictionary-encoded object would have to be handled externally by the callback function, which currently just assumes all the columns will be some type of array. This is still the case with `DictInt*Array`s because a naive transformation of the underlaying array buffer will destroy the relationship between `codes and `codeMapping` but this seems like the less bad failure.

bkmartinjr: I understand that it works fine if you think of these as "an array of codes", but they are defined as "an array of labels", in which case `at()` fails to provide the abstraction.

All I'm really pointing out is that these new API break the abstraction that each column is an array of the actual value type (label, float, whatever).  This is witnessed by the fact that `at()` works sometimes, but not always -- you have to first ask the column "are you a dict encoded column?", and then call the correct accessor.

In this new world, I think you have two choices of abstraction:
* the column is still logically an array of values (not codes, but the actual values)
* or, the column is some new more complex beast which can't be treated like an array
Either works, but it would be good to be clear and build the type abstraction to match.  It will help minimize complexity above this layer.

seve (approved):
Just a couple of questions, that I'd like resolved before merge. But I don't expect a rereview needed.

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,169 @@
...
+    if isinstance(source_array, pd.Index):
+        source_array = source_array.to_series()
+
+    array_class, encoding_dtype = _get_array_class(source_array)
+
+    # two-layer mapper (1) array_class, (2) encoding_dtype
+    # this is necessary as an int32 codes array will require a different coder
+    # than an int32 numeric array.
+    arrayEncoder = {
+        "dense": {
```
atarashansky, line 41: Define this at top-level

server/common/fbs/matrix.py
```diff
@@ -1,23 +1,18 @@
...
 import server.common.fbs.NetEncoding.Matrix as Matrix
-import server.common.fbs.NetEncoding.TypedArray as TypedArray
-import server.common.fbs.NetEncoding.Uint32Array as Uint32Array
+import server.common.fbs.NetEncoding.TypedFBArray as TypedFBArray
 
 
 # Serialization helper
+
+
 def serialize_column(builder, typed_arr):
```
atarashansky, line 16: rename u_type to union_type

server/common/fbs/fbs_coders.py
```diff
@@ -0,0 +1,169 @@
...
+    Coder, array_type = arrayEncoder[array_class].get(encoding_dtype, defaultCoder)
+
+    coder_obj = Coder()
+    # for encoding, we require the source array, flatbuffer builder, and encoding data type
+    array_value = coder_obj.encode_array(source_array, builder, encoding_dtype)
+    return (array_type, array_value)
+
+
+def deserialize_typed_array(tarr):
+    type_map = {
```
atarashansky, line 71: bring to top-level

Bento007 (approved):

Issue #26, opened 2021-08-27
Fix Python linting of PRs


maniarathi:
The push_tests Github Action is supposed to check to make sure that a PR is not ill-formatted before it gets merged into the main branch. Currently, the Github Action first runs `black` to format the files and then runs flake8. Unfortunately this means that the copy of the code that is on the Github Action server is ephemerally "formatted" by black and never fails flake8 even though the PR itself (where black hasn't committed changes) is faulty.

Instead we should be running _checks_ to ensure that the PR adheres to both black standards and flake8 standards. This means that PRs should be running `black directory/path/here --check` and `flake8 directory/path/here` and ensuring that neither returns error codes.

Alternatively, you could make use of an existing Github Action from the marketplace like [this](https://github.com/marketplace/actions/flake8-action). Ensure that we check both black and flake8 though.

Note: this needs to be fixed for both the cellxgene repo and the single-cell-explorer repo.

maniarathi:
Closing this issue directly because the changes are not deployable.

PR #373, opened 2022-08-03
Adjust lint check to also ensure that the formatting matches black format expectations. Also format some files that weren't formatted properly.

maniarathi:
Previously the lint check that we do on all PRs did not check to ensure that the Python code was formatted to `black` standards even though we enforce `black` formatting across all of our repos. This PR adds in that check and also goes back and reformats some files according to the standard that slipped through the cracks because of the missing command in the linting command.

Corresponding issue number: #26

### Reviewers

**Functional:** @atarashansky 

**Readability:** @ebezzi 

---

#### Additions

- Added the `black` linting check to the lint make command which runs on every pull request.

#### Modifications

- Modified the Python backend code to be `black` formatted.

#### Removals

- None.


ebezzi (approved):

atarashansky (approved):
lgtm! thanks arathi!

PR #384, opened 2022-08-10
feature: Added loading spinner to geneset average color-by and fixed a continuous legend bug

### LINKED ISSUE #257: Add loading spinners and clean up unfurling behavior for gene sets
For large datasets when gene sets are slow to load, the user clicks the color by button and then nothing happens, which is confusing. We should add a loading spinner here.

When you unfurl a gene, there's a bunch of extra white space, like its in the expanded state. The gene should remain unexpanded and should have a spinner while loading.

Related: if the user opens a gene set, then closes it and moves on, we don't cancel the fetches.  So it ends up downloading huge amounts of data that then are not used.  In theory, we could cancel the outstanding fetch requests.  In practice, that may be hard to do in our current code base.

atarashansky:
Closes #257 

#### Reviewers
**Functional:** 
@seve @tihuan 

---

## Changes
- added a `colorLoading` variable to the `controls.ts` reducer to control the loading spinner when waiting for color-by.
- updated implementation of the continuous legend to solve the following bug:
  - trigger: clicking on one geneset colorby and then switching to another geneset colorby before the first one loads
  - bad behavior: legend appears when first colorby fetch completes, even though the graph never colors by the first colorby (see below screenshot) image
  - fix: instead of updating with `componentDidUpdate`, use `Async/Fulfilled`

Working demo:
https://user-images.githubusercontent.com/16548075/183977286-28b4618d-9984-4e5f-8e8b-e6fa118ca2fe.mov


tihuan (commented):
LGTM! But I'm not as familiar with Explorer, so maybe wait for Seve for another check 😆 Thank you!

client/src/components/continuousLegend/index.tsx
```diff
@@ -119,51 +120,98 @@ const continuous = (selectorId: any, colorScale: any, colorAccessor: any) => {
   genesets: (state as any).genesets.genesets,
 }))
 class ContinuousLegend extends React.Component {
+  static watchAsync(props: any, prevProps: any) {
+    return !shallowEqual(props.watchProps, prevProps.watchProps);
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
```
tihuan, line 127: Can we type this properly please? In effort to slowly convert Explorer repo to be more typed, we want new code and touched old code to be typed as much as possible 🙏 Happy to help where needed!

context: https://czi-sci.slack.com/archives/C03TR1S2CCU/p1660566509067059

seve (commented):
Does react-async not provide an interface for conditionally updating a component based on the async props? Or did they jump ship on class components? 

I took a preliminary look at the react-async docs, but couldn't find anything on the subject.

atarashansky:
> Does react-async not provide an interface for conditionally updating a component based on the async props? Or did they jump ship on class components?
> 
> I took a preliminary look at the react-async docs, but couldn't find anything on the subject.

I think react-async does provide that interface via `watchProps`, but we weren't using react-async for the continuous legend component.

tihuan (approved):
Amazing!! Thank you so so much for going the extra mile to type the files and improve our code quality ⭐ 👏 !!

PR #386, opened 2022-08-10
fix(navbar): font display

tihuan:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
Update Explorer to use the same font size files as Data Portal, also add global style `-webkit-font-smoothing: antialiased;` per Data Portal 👌 

demo

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -27,12 +27,17 @@
       body {
         margin: 0;
         padding: 0;
+        -webkit-font-smoothing: antialiased;
       }
 
       * {
         box-sizing: border-box;
       }
     </style>
+    <link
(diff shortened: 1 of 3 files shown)


atarashansky (approved):
Thanks for fixing this, @tihuan !!

seve:
@tihuan Why import fonts from Google rather than serve them ourselves?

tihuan:
@seve Great question! My take is that Google's CDN probably already cached fonts at the closest node the user is requesting from, because other users from the same node likely have requested the files already, so the user should get the font faster. Also if user has already downloaded the font from Google from visiting a different site, they don't need to download it again

If we host the font ourselves, we get none of that, plus the overhead of managing how to serve the fonts properly!

PR #390, opened 2022-08-11
Linting with black and prettier - round 2 (forgot `scripts/`)

atarashansky:


tihuan (approved):
rubber stamping given that it's prettier changes

PR #391, opened 2022-08-11
fix: SVG import

tihuan:
Changes:
1. Update webpack `file-loader` to look for all specified file formats in the tree instead of previously restricted `node_modules` and `fonts` folders. This way we can import SVG files anywhere in the project as a file path string
2. Create a basic SvgIcon `img` component to take attribute `src={file_path}`
3. Update `package.json` and `package-lock.json` with `"@material-ui/lab": "^4.0.0-alpha.61",`, otherwise `make build-for-server-dev` would complain about `package.json` and `package-lock.json` are not in sync
4. Remove `publicPath: ".."` in `client/configuration/webpack/webpack.config.prod.js`, since we no longer host our own font files, and we need relative path to load static files like SVG

```
{
        test: /\.(jpg|png|gif|eot|svg|ttf|woff|woff2|otf)$/i,
        loader: "file-loader",
        options: {
          name: "static/assets/[name]-[contenthash].[ext]",
          publicPath: "..", // <--- this is no longer needed
        },
      },
```

Proof that it still works:

Screen Shot 2022-08-11 at 1 41 25 PM

QA:

1. Run `make build-for-server-dev && make start-server` and check out `http://localhost:5005/d/pbmc3k.cxg/`. The SVG icon should work so is the rest of the app!
2. Run `make start-frontend` and check out `http://localhost:3000/d/pbmc3k.cxg`. Everything should work there too 👌


client/src/global.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.svg" {
```
tihuan, line 1: this takes care of TS error

client/src/components/embedding/index.tsx
```diff
@@ -188,8 +77,7 @@ class Embedding extends React.PureComponent<{}, EmbeddingState> {
               <Button
                 type="button"
                 data-testid="layout-choice"
-                icon={heatmapIcon}
-                // minimal
+                icon={<SvgIcon src={heatmapIcon} />}
```
tihuan, line 80: here we render SvgIcon component and pass the SVG file path to it (processed by `file-loader`)

tihuan:
Ah some new lint errors from the minor TS version upgrade in the lock file I think. Will take a look now!

tihuan:
OK lint fixed! Turned out that `package-lock.json` was not having the right ts type version or something, so I did a `rm -rf node_modules && rm package-lock.json && npm i` and that seemed to work!

atarashansky (approved):
Amazing! Thanks so much. Looks great.

PR #423, opened 2022-08-25
chore: typed genesetsUI reducer

atarashansky:


client/src/reducers/genesetsUI.ts
```diff
@@ -1,15 +1,20 @@
+import { AnyAction } from "redux";
+
+export interface GeneSetsUIState {
+  createGenesetModeActive: boolean;
+  isEditingGenesetName: string | false;
+  isAddingGenesToGeneset: string | false;
```
seve, line 6: This seems overloaded, might be smoke showing that we need to refactor something. Could you look into it really quick and create a ticket if necessary.

atarashansky: I don't think they're overloaded - They're used for different things

seve (approved):

PR #424, opened 2022-08-25
chore: typed config reducer

atarashansky:


client/src/reducers/config.ts
```diff
@@ -1,12 +1,24 @@
...
-  state = {
+import { AnyAction } from "redux";
+import type { Config } from "../globals";
+
+interface ConfigState {
+  features: Config["features"] | null;
+  parameters: Config["parameters"] | null;
+  displayNames: Config["displayNames"] | null;
+  loading?: boolean;
+  error?: Error | string | null;
```
seve, line 9: This seems like the result of a downstream bug, shouldn't we restrain ourselves to either Error or string, not both?

seve (approved):

PR #425, opened 2022-08-25
chore: purge autosave from codebase

atarashansky:


client/__tests__/e2e/puppeteerUtils.ts
```diff
@@ -159,9 +159,6 @@ async function nameNewAnnotation() {
   if (await isElementPresent(getTestId("annotation-dialog"))) {
     await typeInto("new-annotation-name", "ignoreE2E");
     await clickOn("submit-annotation");
-
-    // wait for the page to load
-    await waitByClass("autosave-complete");
```
tihuan, line 164: I wonder if we could delete this function completely? It's only used in:

```ts
export async function goToPage(url: any) {
  await page.goto(url, {
    waitUntil: "networkidle0",
  });

  await nameNewAnnotation(); // <---- HERE!
}
```

tihuan (approved):
Great catch on deleting stuff in tests! Just one small question otherwise LGTM 🎉

PR #430, opened 2022-08-26
refactor(cascadeReducer): Quick feedback 2

tihuan:
This is probably the closest we could get to typing actions without using [Redux Toolkit](https://redux.js.org/usage/usage-with-typescript#overview), which is their strongly recommended way to do Redux now. 

The overall pattern is that each slice's reducer's `action` argument is `AnyAction`, since any action goes through every reducer. But in the switch case, we cast the action to the slice's action type, since such action type **should** come with the agreed payload. Note that this is NOT guaranteed at runtime, because Explorer is NOT using action creators to allow TypeScript to map the action type and payload statically. But such refactor would take a lot more time, so this is a compromise for now until we see a strong need to start using Redux Toolkit

### DIFF
--- a/client/src/reducers/datasetMetadata.ts
+++ b/client/src/reducers/datasetMetadata.ts
@@ -1,9 +1,9 @@
 /*
- Dataset metadata reducer, modifies Portal collections-related state. 
+ Dataset metadata reducer, modifies Portal collections-related state.
  */
 
 // Core dependencies
-import { Action } from "redux";
+import { Action, AnyAction } from "redux";
 
 // App dependencies
 import { DatasetMetadata as IDatasetMetadata } from "../common/types/entities";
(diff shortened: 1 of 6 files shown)


client/src/reducers/datasetMetadata.ts
```diff
@@ -1,24 +1,24 @@
...
+import { Action, AnyAction } from "redux";
 
 // App dependencies
 import { DatasetMetadata as IDatasetMetadata } from "../common/types/entities";
 
 /*
  Action dispatched on successful response from dataset-metadata endpoint.
  */
-export interface DatasetMetdataAction extends Action<string> {
+export interface DatasetMetadataAction extends Action<string> {
```
tihuan, line 14: I use https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker to catch typos 🤖 ✅ !

PR #431, opened 2022-08-30
Updating explorer to appropriately handle upgraded CXGs

atarashansky:
#432 

 - The current layout of our gene expression tileDB array (`X`) has been optimized to compromise between fast row slicing and column slicing. However, column slicing for fetching genes is quite slow (~2 seconds for 10 genes in the 1.46m dataset). This PR updates explorer to appropriately handle CXGs with two 1D `X*` arrays, one for row-slicing (`Xr`) and one for column-slicing (`Xc`). The new CXG parameters were loosely (heavy emphasis on the word "loosely") optimized in a Jupyter notebook using the following approach. I kept all other parameters fixed, varied capacity by powers of 2, and picked the capacity that gave the fastest slicing for 10 random indices. Using the new capacity, I did the same for picking the number of tiles. With the new schema, column slicing (i.e. row slicing `Xc`) 10 random genes takes ~80ms. **Note**: CXGs will only have two 1D arrays if the original 2D expression array was sparse. 
 - I also updated the `obs` array to store categorical data as integer codes in the remaster PR and incorporated those changes here. This offers two benefits: (1) integer arrays take up less memory and are faster to load (2) we can avoid having to cast untyped arrays of strings into pandas Categoricals as we already have the codes. The list of categories in the schema type hints (`ObsArray.meta['cxg_schema']`) is ordered to correspond to the integer codes (e.g. for categories ["A","B","C"], 0 in the codes array corresponds to "A", 1 to "B", and so on).

Crucially, I made sure that explorer is backwards-compatible if the CXG has not been remastered yet.

[This PR](https://github.com/chanzuckerberg/single-cell-data-portal/pull/3209) in data portal updates the processing and remastering pipeline to produce the new arrays.


atarashansky:
- [ ] remaster pbmc3k.cxg for unit testing.

client/configuration/webpack/SUPPORTED_BROWSERS_REGEX.js
```diff
@@ -1,2 +1 @@
...
...+(10[_.]3|10[_.]([4-9]|\d{2,})|(1[1-9]|[2-9]\d|\d{3,})[_.]\d+|11[_.]0|11[_.]([1-9]|\d{2,})|11[_.]2|11[_.]([3-9]|\d{2,})|(1[2-9]|[2-9]\d|\d{3,})[_.]\d+|12[_.]0|12[_.]([1-9]|\d{2,})|12[_.]5|12[_.]([6-9]|\d{2,})|(1[3-9]|[2-9]\d|\d{3,})[_.]\d+|13[_.]0|13[_.]([1-9]|\d{2,})|13[_.]7|13[_.]([8-9]|\d{2,})|(1[4-9]|[2-9]\d|\d{3,})[_.]\d+|14[_.]0|14[_.]([1-9]|\d{2,})|14[_.]4|14[_.]([5-9]|\d{2,})|14[_.]8|14[_.](9|\d{2,})|(1[5-9]|[2-9]\d|\d{3,})[_.]\d+|15[_.]0|15[_.]([1-9]|\d{2,})|(1[6-9]|[2-9]\d|\d{3,})[_.]\d+)(?:[_.]\d+)?)|(CFNetwork\/8.* Darwin\/16\.5\.\d+)|(CFNetwork\/8.* Darwin\/16\.6\.\d+)|(CFNetwork\/8.* Darwin\/16\.7\.\d+)|(CFNetwork\/8.* Darwin\/17\.0\.\d+)|(CFNetwork\/8.* Darwin\/17\.2\.\d+)|(CFNetwork\/8.* Darwin\/17\.3\.\d+)|(CFNetwork\/8.* Darwin\/17\.\d+)|(Edge\/(79(?:\.0)?|79(?:\.([1-9]|\d{2,}))?|([8-9]\d|\d{3,})(?:\.\d+)?|83(?:\.0)?|83(?:\.([1-9]|\d{2,}))?|(8[4-9]|9\d|\d{3,})(?:\.\d+)?))|((Chromium|Chrome)\/(61\.0|61\.([1-9]|\d{2,})|(6[2-9]|[7-9]\d|\d{3,})\.\d+|83\.0|83\.([1-9]|\d{2,})|(8[4-9]|9\d|\d{3,})\.\d+)(?:\.\d+)?)|(Firefox\/(60\.0|60\.([1-9]|\d{2,})|(6[1-9]|[7-9]\d|\d{3,})\.\d+)\.\d+)|(Firefox\/(60\.0|60\.([1-9]|\d{2,})|(6[1-9]|[7-9]\d|\d{3,})\.\d+)(pre|[ab]\d+[a-z]*)?)/;
```
atolopko-czi: intended change?

server/compute/diffexp_cxg.py
```diff
@@ -186,7 +181,7 @@ def fp_err_set(err, flag):
 
 
 def mean_var_cnt_dense(matrix, _, rows):
-    xslc = matrix.multi_index[rows, :]
+    xslc = matrix.multi_index[rows]
```
atolopko-czi, line 184: The dense matrix is still 2-dim, but the 2nd dim can be elided, right?

server/dataset/cxg_dataset.py
```diff
@@ -187,11 +196,24 @@ def _validate_and_initialize(self):
         self.about = about
         self.cxg_version = cxg_version
         self.corpora_props = corpora_props
+        self._init_sparsity_and_format()
 
     @staticmethod
     def _open_array(uri, tiledb_ctx):
         return tiledb.open(uri, mode="r", ctx=tiledb_ctx)
 
+    def open_X_array(self, col=False):
```
atolopko-czi, line 205: `col_wise` or `col_major` might be a tad clearer

server/dataset/cxg_dataset.py
```diff
@@ -187,11 +196,24 @@ def _validate_and_initialize(self):
...
+        """
+        A helper function to open the 1D X array in row or column orientation with
+        backwards compatibility for 2D arrays.
+        """
+        if self.is_1d and col:
+            return self.open_array("Xc")
+        elif self.is_1d and not col:
+            return self.open_array("Xr")
+        else:
+            return self.open_array("X")
```
atolopko-czi, line 215: if not using CXG version for conditional, worth noting in a comment here that `X` is deprecated

atolopko-czi (changes requested):
Some minor comments herein. 

A good bit has changed in cxg_dataset.py, and I'm wondering if we have sufficient test coverage for the new Xr/Xc usage obs changes? Existing tests suffice?  I do see a test fixture was updated for pbmc3k.cxg, which I assume is necessary for passing tests.

The cxg.md documents the spec for this file, which should be updated in the PR.

client/configuration/webpack/SUPPORTED_BROWSERS_REGEX.js
atarashansky: No, my local dev stuff keeps updating it...reverted!

server/compute/diffexp_cxg.py
atarashansky: Correct!

atarashansky:
> Some minor comments herein.
> 
> A good bit has changed in cxg_dataset.py, and I'm wondering if we have sufficient test coverage for the new Xr/Xc usage obs changes? Existing tests suffice? I do see a test fixture was updated for pbmc3k.cxg, which I assume is necessary for passing tests.
> 
> The cxg.md documents the spec for this file, which should be updated in the PR.

 - pbmc3k.cxg/X is a dense array so that was not updated. I updated pbmc3k.cxg/obs to store dictionary-encoded categorical data.
 - I included tests for Xr/Xc in data portal. We do not have test coverage for Xr/Xc in explorer because our test fixture is dense data. Would you recommend adding another dataset with sparse values for testing?
 - I will update the documentation for the specs! Thank you for highlighting that.

server/dataset/cxg_dataset.py
atarashansky: X no longer deprecated - it is for dense data. Added a comment!

server/tests/unit/compute/test_diffexp_cxg.py
```diff
@@ -96,6 +96,28 @@ def test_cxg_default(self):
...
+        self.check_1_10_2_10(results)
+
+        # run it directly
+
+        results = diffexp_ttest(adaptor, maskA, maskB, 10)
+        self.check_1_10_2_10(results)
+
+        results = diffexp_ttest(adaptor, maskA.nonzero()[0], maskB.nonzero()[0], 10, selector_lists=True)
+        self.check_1_10_2_10(results)
+
```
atolopko-czi, line 120: won't this clash (be replaced by?) the same-named method below?

server/tests/unit/common/apis/test_api_v3_sparse.py
```diff
@@ -0,0 +1,916 @@
+from urllib.parse import quote
```
atolopko-czi, line 1: I'm not sure if fully duplicating `test_api_v3(_dense).py` is the way to go. Let me give this some thought.

server/tests/unit/compute/test_diffexp_cxg.py
atarashansky: Oh shoot, good catch!!

server/tests/unit/common/apis/test_api_v3_sparse.py
atarashansky: Actually, I think you're right...It's inconvenient to have to add/modify a test in two places

PR #437, opened 2022-09-07
Replace float arrays over the wire with integer coded arrays

atarashansky:
Transmitting dense floats over the wire can result in large response sizes for large datasets. However, the precision afforded by floats is not necessary for an exploratory tool, especially for gene expressions. Histograms only display 40 bins and users will likely be unable to differentiate between more than 100 colors, anyway.

The strategy is to create a new flatbuffer data structure that contains:
 - The digitized float array (N bins) with `int16` precision
 - The maximum value of the source array (`float32`)
 - The minimum value of the source array (`float32`)
 - The number of bins, N

When decoding this flatbuffer, the data is translated back into its float values. 

The nuance is in determining the number of bins. Less bins = less memory. For large datasets, digitizing the layout into too few bins will result in crosshatch appearance after decoding. Empirically, I observed that 5000 bins is granular enough to avoid this visual artifact while reducing memory footprint by more than half. For genes, 500 bins yields histograms, scatterplots, and color-bys that are visually indistinguishable in shape and colors, and reduces memory footprint by ~66%. This was also empirically determined by visual inspection. 

One note: The most accurate way of digitizing (e.g. via `numpy.histogram` and `numpy.digitize`) is a costly operation (> 200ms for >1m cells). A 10X faster approach is to divide the source array by the max value, multiply by the number of bins, and perform element-wise integer truncation. This approximation results in <1% of the resulting digits erroneously shifted by 1 bin. However, for 500 bins this amounts to 0.2% error in the decoded float value compared to accurate digitization. 

Some quantitative validation:
Average RMSE (decoded float array after lossy encoding vs source float array) of 50 random genes on the 1.46m dataset vs number of bins. Red line is drawn at #bins=500.
image
pbmc3k data:
image

Average RMSE of the X and Y coordinates for the t-SNE in the 1.46m dataset. Red line is drawn at #bins=5000.
image
pbmc3k data:
image

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- added a new flatbuffer type for encoding floats as integer for smaller over-the-wire memory footprints


atarashansky:
- [ ] Add unit tests for new flatbuffer type for varying number of bins.

fbs/matrix.fbs
```diff
@@ -62,6 +63,17 @@ table DictEncoded32FBArray {
...
+  // the integer codes corresponding to float bins
+  codes: [int16];
+  // the maximum value of the source array
+  max: float32;
+  // the minimum value of the source array
+  min: float32;
+  // the number of bins
+  nbins: int32;
+}
+
```
atolopko-czi, line 76: worth spelling out here how the encoding works, maybe just via a single math expression (repeating [this](https://github.com/chanzuckerberg/single-cell-explorer/pull/437/files#diff-0c1e70ff198b19b032eaf0c7e925217a2bfa6e135bdd6eab94f095bdcfd6b2bfR52))

fbs/matrix.fbs
```diff
@@ -9,6 +9,7 @@
     - JSON/UTF8 encoded array (for other types)
     - signed and unsigned 8, 16, 32, and 64 bit integers
     - dictionary-encoded categorical arrays with 8, 16, and 32 bit integer codes
+    - integer-encoded float arrays
```
atolopko-czi, line 12:
```suggestion
    - integer-encoded float arrays (lossy)
```

server/common/fbs/fbs_coders.py
```diff
@@ -13,12 +13,49 @@
...
+
+DEFAULT_NUM_BINS = 5000
+
+
+class DenseNumericIntCoder:
+    n_slots = 4
+
+    def encode_array(self, array, builder, _dtype, num_bins=None):
+        if num_bins is None:
+            num_bins = DEFAULT_NUM_BINS
```
atolopko-czi, line 26: is there ever a case where is this gets called without an explicit `num_bins` arg? Seems safer to error if so and force callers to specify the bin count.

client/src/util/stateManager/matrix.ts
```diff
@@ -22,6 +22,7 @@ import { Matrix } from "./net-encoding/matrix";
 import { Column } from "./net-encoding/column";
 
 const utf8Decoder = new TextDecoder("utf-8");
+const DEFAULT_NUM_BINS = 5000;
```
atolopko-czi, line 25: ideally this value would be communicated via the server config so we can ensure it stays in sync with the server encoding

atolopko-czi: Actually I see it's only used for encoding here, but why do we ever have to encode on the client side?

client/__tests__/e2e/__snapshots__/e2e.test.ts.snap
```diff
@@ -2,8 +2,8 @@
...
...\"fill: black;\\"></rect><rect x=\\"17.500000000000004\\" y=\\"14.913093858632676\\" width=\\"2.7499999999999964\\" height=\\"0.086906141367324\\" style=\\"fill: black;\\"></rect><rect x=\\"20.25\\" y=\\"14.965237543453071\\" width=\\"2.75\\" height=\\"0.03476245654692889\\" style=\\"fill: black;\\"></rect><rect x=\\"23\\" y=\\"14.913093858632676\\" width=\\"2.7500000000000036\\" height=\\"0.086906141367324\\" style=\\"fill: black;\\"></rect><rect x=\\"25.750000000000004\\" y=\\"14.80880648899189\\" width=\\"2.7499999999999964\\" height=\\"0.19119351100811066\\" style=\\"fill: black;\\"></rect><rect x=\\"28.5\\" y=\\"14.652375434530708\\" width=\\"2.7500000000000036\\" height=\\"0.34762456546929243\\" style=\\"fill: black;\\"></rect><rect x=\\"31.250000000000004\\" y=\\"14.47856315179606\\" width=\\"2.7500000000000036\\" height=\\"0.5214368482039404\\" style=\\"fill: black;\\"></rect><rect x=\\"34.00000000000001\\" y=\\"14.530706836616455\\" width=\\"2.75\\" height=\\"0.4692931633835453\\" style=\\"fill: black;\\"></rect><rect x=\\"36.75000000000001\\" y=\\"14.25260718424102\\" width=\\"2.749999999999993\\" height=\\"0.74739281575898\\" style=\\"fill: black;\\"></rect><rect x=\\"39
```
atolopko-czi, line 7: hard to see the diff; is this effectively changing the expected result for a test?

atolopko-czi (changes requested):
Mostly just Q's, which may highlight my own misunderstandings. Overall strategy for encoding/compression LGTM!

One overall readability concern is that it's hard to tell what client-requested data is getting into encoded and what the bin counts are. Mostly because the reader has to follow the call hierarchy to see where the bin counts are specified.  I don't think it's a huge deal but maybe it's worth documenting at the API level which API calls are affected by this new encoding.

client/__tests__/e2e/__snapshots__/e2e.test.ts.snap
atarashansky: Correct, this is updating the snapshot puppeteer uses to make sure the UI looks as expected after performing an action.

client/src/util/stateManager/matrix.ts
atarashansky: We used to encode on the client side when we allowed users to annotate the data (in desktop). I think we are planning at some point in the future to add that back to hosted explorer.

server/tests/unit/dataset/test_cxg_dataset.py
```diff
@@ -47,4 +47,4 @@ def test_tdb_bug(self):
         self.assertIs(type(summary["columns"]), list)
         self.assertEqual(len(summary["columns"]), 1)
         self.assertEqual(len(summary["columns"][0]), 2638)
-        self.assertEqual(summary["columns"][0].sum(), np.float32(-19.00301))
+        self.assertEqual(summary["columns"][0].sum(), np.float32(-35.90116))
```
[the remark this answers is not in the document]
atarashansky: This is correct, but I was also curious about why the change was so significant. The takeaway is that deviations in per-element float values which are individually small amount to a large difference when taking the sum across all elements. 

For example, the mean of the source array is `-0.0072` and the mean of the decoded(encoded(source array)) is `-0.0136`, both of which are really small compared to the range of expression values in the array. These particular genes are really sparse, so after standardization there are a ton of small negative values which leads to this effect.

This plot shows the the magnitude of the source array vs the % deviation in magnitude between source and lossy arrays (`|source-lossy| / |source| * 100`):
image
The largest deviants are all (negligibly) small values that, when summed up, give a significantly different sum.

atarashansky:
> One overall readability concern is that it's hard to tell what client-requested data is getting encoded and what the bin counts are. Mostly because the reader has to follow the call hierarchy to see where the bin counts are specified. I don't think it's a huge deal, but maybe it's worth documenting at the API level which API calls are affected by this new encoding.

Great point. I think I'm going to add the bin counts as an optional parameter to the get requests for `layout/obs`, `data/var` and `summarize/var` requests. If not specified, no encoding is used. If specified, we use the encoding. That way, default value don't need to be hardcoded in two separate places (BE and FE) and the BE makes no assumptions about how the client wants its data returned.

atarashansky:
@atolopko-czi I decided to go through with it and build it into the API. On the client side, I needed to add an optional `nBins` argument to the AnnoMatrix.`fetch` interface and downstream functions. I initially wanted to create a more complex query type for `obs`, `var`, and `emb` queries that would include `nBins` and leave the `fetch` interface unchanged, but that started getting even more complicated.

server/common/fbs/fbs_coders.py
```diff
@@ -147,6 +146,11 @@ def decode_array(self, u, TarType):
         np.dtype(np.uint32).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
         np.dtype(np.uint64).str: (DenseNumericCoder, TypedFBArray.TypedFBArray.Uint32FBArray),
     },
+    "lossy": {
+        np.dtype(np.float64).str: (DenseNumericIntCoder, TypedFBArray.TypedFBArray.Int16EncodedXFBArray),
+        np.dtype(np.float32).str: (DenseNumericIntCoder, TypedFBArray.TypedFBArray.Int16EncodedXFBArray),
+        np.dtype(np.float16).str: (DenseNumericIntCoder, TypedFBArray.TypedFBArray.Int16EncodedXFBArray),
+    },
```
atolopko-czi, line 153: nice, i like `lossy` key/group

atolopko-czi (approved):
Pushing the `num_bins` values all the way to the client side is a nice solution. Thanks for the additional tests for "lossy" encoding. Two minor doc-related  comments for you to see before you merge. Otherwise LGTM!

Issue #209, opened 2022-01-12
Upgrade Flask package dependency to be > 2.0


maniarathi:
We previously could not upgrade Flask to the major new release (2.0) because we also had a dependency on Flask-restful which was not compatible with the new major version. Flask-Restful now has a new version, 0.3.9, that is compatible with Flask 2.0 so we should upgrade both. 

Please make sure to also update the requirements in single-cell-infra as well here: `terraform/modules/hosted-cellxgene/common/customize/requirements.txt`.

maniarathi:
This issue needs to be updated to be generalized to update all Python dependencies (including Flask) on the Explorer.

atolopko-czi:
associated PR is merged; both DP & Explorer have Flask >2.0.. Closing

PR #438, opened 2022-09-12
Upgrade Flask and other server python dependencies

mlin:
#209

To upgrade Flask & Flask-RESTful, (i) removing pins in a couple of other packages and (ii) accommodating a minor API change (constraining the internal names of Blueprints), was sufficient to make the server unit+smoke tests pass.

I could use advice on additional testing we might want/need in order to have more confidence in these changes. My latent worry is that some of the removed pins were avoiding known problems with subsequent versions of those dependencies, and whether we have regression tests for those issues.


.github/workflows/push_tests.yml
```diff
@@ -3,7 +3,7 @@ name: Push Tests
 on:
   push:
     branches:
-      - main
+      - "**"
```
mlin, line 6: Is this OK? I think it's useful to have WIP branches automatically tested even if they're not yet a PR, but LMK if there's some history here

atarashansky: I can't think of any reason it was set to main other than not wanting to fill up the GHA queue with a ton of tests unrelated to active PRs. I think changing to all branches is fine for now, and if there's pushback we can revert.

server/app/api/v3.py
```diff
@@ -199,7 +199,7 @@ def register_api_v3(app, app_config, server_config, api_url_prefix):
 
     s3uri_api_path = "s3_uri"
     bp_s3uri = Blueprint(
-        f"api_dataset_{s3uri_api_path}_{api_version}",
+        f"api_dataset_{s3uri_api_path}_{api_version.replace('.',',')}",
```
atarashansky, line 202: Is `.` a reserved character or something?

.github/workflows/push_tests.yml
atarashansky: image
^ 10 checks now

atarashansky (changes requested):
I think the `main`-->`**` change might be resulting in duplicate tests being run. Otherwise, changes look good!

server/app/api/v3.py
mlin: Yea, in Flask 2 they became ineligible to use in the blueprint name -- something about [namespacing for nested blueprints](https://github.com/pallets/flask/issues/4041). As far as I could tell the blueprint name is just an internal handle that shows up in logs etc. but doesn't have any observable effect on the API server. Happy to be corrected if that's not true...

mlin:
OK, reverted the branch testing change. I'd still like to do more testing on this though -- suggestions? open it up in-browser and click on stuff until something breaks? (or does the "e2e" testing do something like that?)

atarashansky:
> OK, reverted the branch testing change. I'd still like to do more testing on this though -- suggestions? open it up in-browser and click on stuff until something breaks? (or does the "e2e" testing do something like that?)

The smoke tests should have pretty good coverage. I'll open it up in browser and try to break it! Will approve if all looks good.

atarashansky:
It all looks good!

atarashansky (approved):

Issue #455, opened 2022-09-20
Improvement: Add smoke test coverage for continuous clipping with categorical metadata open


atarashansky:
The bug indicated in #453 failed to be caught by smoke tests, because the smoke tests did not check if categorical labels are correct when continuous clipping is active.

[This test](https://github.com/chanzuckerberg/single-cell-explorer/blob/92c5237a9ad9bcb2925a9494652a5f84991c82e6/client/__tests__/e2e/e2e.test.ts#L262) should add actions to open the `louvain` category and confirm the labels are correct.

PR #456, opened 2022-09-20
chore: Update clipping smoke test to ensure categorical data appears properly when clipping

atarashansky:
#455


atolopko-czi (approved):
Looks reasonable!  Only suggestion is that this could be broken out as a separate test, so we have separate continuous and categorical value tests; particularly considering the current test name.

ashin-czi (approved):

PR #462, opened 2022-09-24
fix: fixing a bug that occurs when there are more than 1000 categories

atarashansky:
Context: Cellxgene purposefully truncates categorical data when there are more than 1000 categories (i.e. it keeps the top 1000 most abundant categories and replaces all other categories with a dummy value). 

With the new dictionary-encoded data structure introduced in the frontend came a bug that occurs when there are more than 1000 categories. The issue was that the dictionary encoded array was converted into an integer array when creating a new metadata column with truncated labels. Now, we have a check to make sure dictionary-encoded arrays are converted into value arrays for the truncated metadata column.


client/__tests__/e2e/e2e.test.ts
```diff
@@ -90,12 +90,17 @@ const genesetDescriptionString = "fourth_gene_set: fourth description";
...
+const dataTruncate = datasets[DATASET_TRUNCATE];
 
 const defaultBaseUrl = "d";
 const pageUrl = appUrlBase.includes("localhost")
   ? [appUrlBase, defaultBaseUrl, DATASET].join("/")
   : appUrlBase;
 
+  const pageUrlTruncate = appUrlBase.includes("localhost")
+  ? [appUrlBase, defaultBaseUrl, DATASET_TRUNCATE].join("/")
+  : appUrlBase;
```
seve, line 102: I see it does it above as well, but why does this default to url base if it's not localhost?

client/src/annoMatrix/normalize.ts
```diff
@@ -163,13 +164,17 @@ export function normalizeCategorical(
   topNCategories.add(overflowCatName);
 
   // rewrite data - consolidate all excess labels into overflow label
-  const newColData = Array.from(col.asArray());
+  const colData = col.asArray();
+  const newColData = new Array(colData.length); // Array.from(col.asArray());
```
seve, line 168:
```suggestion
  const newColData = new Array(colData.length);
```

seve (approved):
Small question + nit

client/__tests__/e2e/e2e.test.ts
atarashansky: Good question

Issue #478, opened 2022-09-29
chore: Add smoke test for dataset breadcrumbs


atarashansky:
I have attempted to add the proper `dataset-metadata` and collection home link stubs (see commented section for the latter [here](https://github.com/chanzuckerberg/single-cell-explorer/blob/556da57600df31ba0410442eb69c3e037191ae23/client/__tests__/e2e/puppeteer.setup.ts#L43)). I have also added test IDs to the breadcrumbs elements. Unfortunately, these components are still not appearing in puppeteer.

Issue #460, opened 2022-09-23
bug: Add feature flag to hide unified nav bar on datasets with invalid `/dataset-metadata` response.


atarashansky:
Remove the unified nav bar for covid 19 cell atlas datasets by adding an additional flag to the Explorer config and disallow the d dataroot.

Example: https://cellxgene.cziscience.com/d/cellular_census_of_human_lungs_alveoli_and_parenchyma-15.cxg/
[Relevant slack thread](https://czi-sci.slack.com/archives/C03PT4ZN46A/p1663896183753079)

PR #482, opened 2022-09-30
fix: hide navbar when seamless not enabled

atarashansky:
#460 
#478 

Some datasets (e.g. on the covid19 atlas portal) are living on CZI infra but not hosted by the data portal. These should not have the unified navigation bar displayed.

Data portal datasets should all have valid results from the `dataset-metadata` endpoint. As a result, any dataset with an invalid `dataset-metadata` result will have their nav bars hidden. 

To ensure tests work properly, dataset metadata and the collections home link page needed to be stubbed. As a result, breadcrumbs smoke tests can now work, so I added them to this PR as well (#478).


client/__tests__/e2e/puppeteer.setup.ts
```diff
@@ -5,12 +5,11 @@
  */
 
 import { setDefaultOptions } from "expect-puppeteer";
+import fetch from "puppeteer-fetch";
 import { isDebug, isDev } from "./config";
 import * as ENV_DEFAULT from "../../../environment.default.json";
 import { DATASET_METADATA_RESPONSE } from "../__mocks__/apiMock";
-
 // TODO: uncomment this for breadcrumbs tests
```
tihuan, line 12: should we remove this comment, too, now that we always test breadcrumbs?

Thank you!

tihuan (approved):
LGTM! Thanks so much for the fix, Alec 🙏 !

PR #489, opened 2022-10-06
refactor: remove v2

### LINKED ISSUE #208: Remove v2 API support
Now that we have v3 API in use in production, we can safely remove the[v2 API class](https://github.com/chanzuckerberg/single-cell-explorer/blob/020b2ffe76cc449d165d3ce43e56dfad020b66d8/server/app/api/v2.py#L1-L0)from Explorer. Note [this todo](https://github.com/chanzuckerberg/single-cell-explorer/blob/020b2ffe76cc449d165d3ce43e56dfad020b66d8/server/app/api/v2.py#L63).

Bento007:
#### Reviewers
**Functional:** @danieljhegeman 

**Readability:** @atolopko-czi 

---

## Changes
- fix #208
- remove v2 code and tests
- move dataset_index out of server.app
- Updating tests to work with V0.3
- Update Black to a compatible version.
- rename rest_get_data_adapter for v0.3 to rest_get_s3uri_data_adapter to better describe it's purpose.
- add rest_get_dataset_explorer_location_data_adaptor to v3 which was rest_get_data_adapter copied from v2. 
- Move DatasetMetadataAPI from v2 to v3
- Remove unused code from v3 register_api_v3. We should be able to close https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell-explorer/110 
- Add after request logging to the server app.


server/tests/unit/common/config/test_dataset_config.py
```diff
@@ -119,19 +120,28 @@ def test_multi_dataset(self):
...
+        def _get_v03_url(url):
+            response = session.get(f"{url}/api/v0.3/s3_uri")
+            s3_uri = quote(quote(response.json, safe=""), safe="")
+            return f"/s3_uri/{s3_uri}/api/v0.3"
+
+
         with self.subTest("Test config for dataroot /set1/1/2/ returns the s1 config"):
-            response1 = session.get("/set1/1/2/pbmc3k.cxg/api/v0.2/config")
+            v03_url = _get_v03_url("/set1/1/2/pbmc3k.cxg")
+            response1 = session.get(f"{v03_url}/config")
```
atolopko-czi, line 131: It's preferable to specify full literal values in tests, for readability.  Being DRY here and/or computing test-related values isn't removing much code, but it is forcing reader to skip around.

If we're only testing the methods that take an s3_uri, and not testing the s3_uri lookup api call it self, I would make it clearer in these tests by doing the s3_uri lookup in a setup method (or commented section), or even just hardcode the s3_uri (it's deterministic). If we _are_ testing the s3_uri lookup, I wouldn't relegate it to a method; I would keep it inlined in the test. Or test the s3_uri lookup directly in its own test.

atolopko-czi (approved):
LGTM! I did a suggestion about some test code; see what you think.  

A secondary question I had is whether it is even worth keeping a v3 designation at all? The server and client are deployed together, so stay in lock step. API versioning only helps when we have users with active Explorer sessions that we don't want to disrupt. But any future backward incompatibilities that we might introduce can still be addressed an API _method_ level, rather than having full API versioning.

server/tests/unit/common/config/test_dataset_config.py
Bento007: hard coding is difficult since it's the full path to the fixture. I'll add comments

Issue #181, opened 2021-11-01
Remove "single dataset" mode


atolopko-czi:
Before the split of Explorer into separate desktop and hosted repos, the code supported both "single" and "multi-root" dataset modes. Only that latter is needed for hosted explorer. Remove all "single dataset" code, for which there is considerable config-related logic.  (search for "single_dataset").

Worth doing _before_ any config system redesign work, since it may lessen the footprint of the code that needs to be grokked and redesigned.

DoD:
* All configuration logic and default config properties related to "single dataset" mode have been removed
* Explorer works exclusively in "multi-root dataset" mode. All tests pass (unit and E2E).
* RDS postgreql db for Explorer is deprovisioned via single-cell-infra changes.
* Dev documentation updated to remove any mention of "user annotations" or "labels"
* Team informed of change.

danieljhegeman:
Worth noting that removing 'single dataset mode' code does _not_ include removing all/any (I need to go back and look) of the cli code; the cli is used for launching the local development server.

metakuni:
Re: `RDS postgreql db for Explorer is deprovisioned via single-cell-infra changes.` under DoD, see also #1276

PR #495, opened 2022-10-12
refactor: single dataset mode

Bento007:
- #181
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- Remove single dataset mode parameters from config and update comments
- Remove single dataset mode if conditions
- multi-dataset mode is required in the config. An error is raised if it's not configured.
- Removing unused test functions
- Remove single dataset mode tests


server/dataset/dataset_metadata.py
```diff
@@ -34,11 +34,8 @@ def infer_dataset_s3_uri(server_config: ServerConfig, dataset_root: str, dataset
     Use the dataset_root, dataset_id, and the server config to infer the physical S3 URI for an Explorer dataset
     artifact
     """
-    # TODO @mdunitz remove after fork, update config to remove single_dataset option, the multiroot lookup will need to
-    #  remain while we support covid 19 cell atlas.
+    # TODO @mdunitz remove after fork, the multiroot lookup will need to remain while we support covid 19 cell atlas.
```
metakuni, line 37: I guess this TODO (`@mdunitz`) will never get done 😆

server/common/config/server_config.py
```diff
@@ -143,7 +135,7 @@ def handle_data_locator(self):
                 for path in paths:
                     if path.startswith("s3://"):
                         break
-            if path.startswith("s3://"):
+            if isinstance(path, str) and path.startswith("s3://"):
```
atolopko-czi, line 138: bug fix?

server/tests/unit/__init__.py
```diff
@@ -3,43 +3,23 @@
...
-
-        return wraps
-
-    return decorator
-
-
-def app_config(data_locator, extra_server_config={}, extra_dataset_config={}):
+def app_config(extra_server_config=None, extra_dataset_config=None):
+    extra_server_config = extra_server_config if extra_server_config else {}
+    extra_dataset_config = extra_dataset_config if extra_dataset_config else {}
```
atolopko-czi, line 16:
```suggestion
    extra_dataset_config = extra_dataset_config or {}
```

atolopko-czi: curious why these new lines are now needed?

atolopko-czi (changes requested):
All of the removals look reasonable. (Admittedly, it's hard to see the big picture w/o a deeper code dive.) 

Before approving, what do you think of removing the term "multi_dataset" everywhere, now that we no longer need to differentiate between the single/multi modes?

server/common/config/server_config.py
Bento007: yes, in some tests the path is None.

server/dataset/dataset_metadata.py
Bento007, line 37:
```suggestion
```

server/tests/unit/__init__.py
Bento007: those parameters were default set to `{}` which is mutable and could cause strange behavior if manipulated. I changed the defaults to `None` which is immutable, and set them to `{}` if `None`

Bento007:
I think removing `multi_dataset` would be a mistake since the name is used to parse the config, and there are setting nested under the mult_dataset setting. Changing the name is an option, but I can't think of a better name. When we clean up the config, a better name might become easier to decide on or obsolete.

server/dataset/dataset_metadata.py
```diff
@@ -34,11 +34,7 @@ def infer_dataset_s3_uri(server_config: ServerConfig, dataset_root: str, dataset
     Use the dataset_root, dataset_id, and the server config to infer the physical S3 URI for an Explorer dataset
     artifact
     """
-    # TODO @mdunitz remove after fork, update config to remove single_dataset option, the multiroot lookup will need to
-    #  remain while we support covid 19 cell atlas.
     #   See ticket https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/corpora-data-portal/1281 # noqa
```
atolopko-czi, line 37: Not sure that this issue reference is useful to understanding. Remove? If should keep, at least update to the correct ticket (it moved repos):
```
    #   See ticket https://github.com/chanzuckerberg/single-cell-explorer/issues/119 # noqa
```

atolopko-czi (approved):

Issue #119, opened 2021-10-07
Clean up config


maniarathi:
**Need**: the current config has probably too many different knobs that also interact with one another. This has caused developer difficulty in the past as the config is hard to understand and difficult to modify.

**Approach**:
- Hard code any vars specific to the cellxgene Explorer
- Remove any vars specific to the cellxgene Desktop
- Consolidate the different configuration modules
- Clean up tests
- Update documentation
~- Remove support for single dataset~
- Update config to accept yaml file
- Do type checking via the yaml file, remove type checking code from config validation process
- Update configs in single-cell-infra
~- remove support for experimental features~
~- remove support for annotations ontology file~

**Definition of Done**: the config is approachable and understandable for a new engineer or user and only contains values that we need to configure.

PR #500, opened 2022-10-13
Simplify Config

Bento007:
#119 

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- Merge `server_config` and `dataset_config` into `app_config`. This adds more complexity than needed.
- Validate `app_config` using pydantic. This greatly simplifies the config validation processes and uses existing python type hinting to do it. 
- modify `AppConfig.__init__` to accept a file name as a parameter.
- Remove `external_config` and `secret_config`. This added more complexity that we weren't really using. Pydnatic support this functionality and we can add it back in the future if needed.
- Remove getting config from environment variables. This is part of removing the external config. 
- Remove the `external` section from config. This is part of removing the external config. 
- Rename the `dataset` config parameter to `default_dataset`. This is a better description of what it is.
- Add `dataroots` config parameter where multiple dataroots can be specified. To simplify the requirement that `dataroot` can be both a string and dictionary, `dataroots` was added to contain the dictionary version of `dataroot`. This simplifies the configuration and validation logic and adds minimal overhead.
- Rename `s3: region_name` config parameter to `s3_region_name`. This simplifies the config by reducing the nesting level for a single key dictionary.
- Rename `MatrixDataLoader` to `DataLoader`. This was a TODO in the code.
- Use `current_app.app_config` in the API handlers rather than passing `app_config` as a parameter. This reduces the complexity of the interface for some of the API handlers.
- Remove `per_dataset_config` because it's not used and adds significant complexity.
- Remove `allowed_matrix_types` because only CXG is supported.
- Remove `get_title` and `get_about` from `app_config`. All it does is wrap the equivalent data_adapter function.
- Remove obsolete tests.
## Testing
Checked that it works on rdev

---
## Future
- Recommend moving documentation to the AppConfigModel.py
- remove default_config.py


server/cli/launch.py
```diff
@@ -232,52 +231,35 @@ def launch(
...
-        cli_config.update_server_config(
-            app__verbose=verbose,
-            app__debug=debug,
-            app__host=host,
-            app__port=port,
-            app__open_browser=open_browser,
-            multi_dataset__dataroot=dataroot,
+        updates = dict(
+            server__app__verbose=verbose,
+            sserver__app__debug=debug,
```
Bento007, line 242:
```suggestion
            server__app__debug=debug,
```

ebezzi (approved):

atarashansky:
@Bento007 @ebezzi this PR broke local development.

PR #502, opened 2022-10-17
fix: Assume NavBar is present when calculating height of explorer initially.

atarashansky:
https://github.com/chanzuckerberg/single-cell-explorer/issues/501


client/src/components/framework/layout.tsx
```diff
@@ -30,7 +30,7 @@ const Layout: React.FC<Props> = (props) => {
     <div
       style={{
         display: "grid",
-        paddingTop: seamlessEnabled ? globals.HEADER_HEIGHT_PX : 0,
+        paddingTop: seamlessEnabled ?? true ? globals.HEADER_HEIGHT_PX : 0,
```
tihuan, line 33: can we add a comment here to explain the context why null coalescing operator `??` is needed please?

e.g.,

```ts
// `??` is used so when `seamlessEnabled` is `undefined` or `true`, so we can add padding to prevent 
// unneeded scrollbar
paddingTop: seamlessEnabled ?? true ? globals.HEADER_HEIGHT_PX : 0,
```

tihuan (approved):
LGTM! Just a quick comment 👌 Thanks for fixing this!

client/src/components/framework/layout.tsx
tihuan: I probably need to look into this a little more, but so far I can't tell why this fixes the issue in the sense that while loading `seamlessEnabled` is `undefined`, so this div doesn't get `paddingTop`. But once `seamlessEnabled` is `true`, it should/will re-render and add the padding, so we don't need `??` to add padding while it's loading?

I'm probably missing something 😆

atarashansky: I'm also not entirely sure if this fixes the issue, as I couldn't reproduce the issue locally...

We diagnosed the issue to be the following:
1. Client loads, NavBar is still missing.
2. Viewport height is calculated.
3. NavBar appears after valid endpoint response.
4. Viewport height is not recalculated until a resize happens.

The problem is that I'm not sure why the viewport height isn't recalculated until resize occurs...

PR #508, opened 2022-10-19
fix: remove eb from Dockerfile

danieljhegeman:
#### Reviewers
**Functional:** 
@Bento007 
**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/hosted/Dockerfile
+++ b/hosted/Dockerfile
@@ -28,9 +28,6 @@ RUN make build-client
 
 ADD server/ server/
 RUN mkdir -p /server/common/web/templates/
-
-WORKDIR server/eb
-RUN make build-ecs
 WORKDIR /
 
 ADD hosted/config.yaml config.yaml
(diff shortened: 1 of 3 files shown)


hosted/Dockerfile
```diff
@@ -28,9 +28,6 @@ RUN make build-client
 
 ADD server/ server/
 RUN mkdir -p /server/common/web/templates/
-
-WORKDIR server/eb
-RUN make build-ecs
```
atolopko-czi, line 33: Are you sure we don't need this? (note it's ECS, not EB).

Bento007 (changes requested):
We will need to approach this more surgically https://github.com/chanzuckerberg/single-cell-explorer/pull/509

danieljhegeman:
> We will need to approach this more surgically #509

Fine by me. Closing.

PR #503, opened 2022-10-17
refactor: Remove elastic bean stock path

Bento007:
#### Reviewers
**Functional:** @atolopko-czi 

**Readability:** @danieljhegeman 

---

## Changes
- remove elastic bean stock path

PR #509, opened 2022-10-19
Revert "refactor: Remove elastic bean stock path"

Bento007:
Reverts chanzuckerberg/single-cell-explorer#503


atolopko-czi (approved):
sorry this became harder than originally thought--but this makes it even more valuable to do, IMO.

PR #518, opened 2022-10-24
refactor: remove eb files

Bento007:
#### Reviewers
**Functional:** @ebezzi 

**Readability:** @atolopko-czi 

---

## Changes
- move make recipe `build-ecs` to ./server/makefile
- remove files in ./server/eb
- modify docker and make files that reference the `eb` path


atolopko-czi (approved):
💯 Thanks for taking a second pass at this one! `--tech_debt_score`!

PR #351, opened 2022-07-27
fix: gene info QA changes

sienacizdziel:
#### Reviewers
**Functional:** 
@seve 
**Readability:** 

---

## Changes
- fix gene info card hiding feature in controls.ts
- remove gene info card if info button is clicked while card is open
- change all instances of "Roboto Condensed" to "Inter"
- download and import "Inter" font family
- disable info button while genesets load
- move gene info analytics to separate branch for review
- if mini histogram is not visible or loading, keep gene info button on the left
- fix histogram loading height

## Notes for Reviewer
Link to QA doc [here](https://docs.google.com/document/d/1TZgUXv9QBAshJCYSjnS8Qs6bgeZ6u48awKYAkUoDzaI/edit)

PR #520, opened 2022-10-25
refactor: Don't Populate __init__.py

ashin-czi:
#351

#### Reviewers
**Functional:** @danieljhegeman @Bento007 @atolopko-czi 

**Readability:** @danieljhegeman @Bento007 @atolopko-czi 

---

## Changes
- add
- remove
- modify
    - Removing code out of __init__.py and placing them in other files


server/common/utils/cache.py
```diff
@@ -0,0 +1,41 @@
+from functools import wraps
```
atolopko-czi, line 1: perhaps name this module `http_cache.py`

server/common/utils/version.py
```diff
@@ -0,0 +1,2 @@
+__version__ = "0.16.7"
+display_version = "cellxgene v" + __version__
```
atolopko-czi, line 2: It would be acceptable to have this file all that way up in the `backend` dir.   I wouldn't consider it a `util` module.

atolopko-czi (approved):
👍 , with 2 minor suggestions

server/common/utils/version.py
Bento007: Make this change before committing

Bento007 (changes requested):

atolopko-czi (approved):
LGTM!

Bento007 (approved):

PR #545, opened 2022-12-13
feat(wmg): 3701 Remove beta tag

tihuan:


tihuan:
I actually couldn't get `make build-for-server-dev && make start-server` to work locally. Got the error that `pydantic` doesn't exist, but `make build-server` didn't work either. So this PR is flying blind a little, but pretty low risk, so will test in `dev` 😆 Thank you!

atarashansky (approved):

PR #577, opened 2023-03-23
feat: increase zoom limits by a factor of 4

atarashansky:
## Changes
- increase zoom limit to 12x from 3x
- add smoke test for zoom
- prevent scrolling on graph component

## Testing
 - tested locally and smoke tested

## Notes for reviewer
Previously, scrolling behavior conflicts with zoom tool:
https://user-images.githubusercontent.com/16548075/231508562-a850bf9c-8861-442f-bdbc-122236288540.mov

Now, the rubberbanding scrolling behavior is no longer present on the graph component:
https://user-images.githubusercontent.com/16548075/231508671-2467c602-80b2-4d17-afd9-42eee2b294a2.mov

### DIFF
--- a/client/__tests__/e2e/cellxgeneActions.ts
+++ b/client/__tests__/e2e/cellxgeneActions.ts
@@ -46,15 +46,12 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
 export async function scroll({testId, deltaY, coords}: {testId: string, deltaY: number, coords: number[]}) {
   const layout = await waitByID(testId);
   if (layout){
-    const elBox = await layout.boxModel();
-    if (elBox) {
-      const x = coords[0];
-      const y = coords[1];
-      await page.mouse.move(x, y);
-      await page.mouse.down();
-      await page.mouse.up();
-      await page.mouse.wheel({ deltaY });
-    }
+    const x = coords[0];
+    const y = coords[1];
+    await page.mouse.move(x, y);
+    await page.mouse.down();
+    await page.mouse.up();
+    await page.mouse.wheel({ deltaY });
   }
 }
 
--- a/client/__tests__/e2e/cellxgeneActions.ts
+++ b/client/__tests__/e2e/cellxgeneActions.ts
@@ -43,6 +43,21 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
   await page.mouse.up();
 }
 
+export async function scroll({testId, deltaY, coords}: {testId: string, deltaY: number, coords: number[]}) {
+  const layout = await waitByID(testId);
+  if (layout){
+    const elBox = await layout.boxModel();
+    if (elBox) {
+      const x = coords[0];
+      const y = coords[1];
+      await page.mouse.move(x, y);
+      await page.mouse.down();
+      await page.mouse.up();
+      await page.mouse.wheel({ deltaY });
+    }
+  }
+}
+
 export async function keyboardUndo() {
   await page.keyboard.down("MetaLeft");
   await page.keyboard.press("KeyZ");
--- a/client/__tests__/e2e/e2e.test.ts
+++ b/client/__tests__/e2e/e2e.test.ts
@@ -26,6 +26,7 @@ import {
   calcDragCoordinates,
   calcTransformDragCoordinates,
   drag,
+  scroll,
   expandCategory,
   subset,
   createGeneset,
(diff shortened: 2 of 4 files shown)


client/src/components/graph/graph.tsx
```diff
@@ -918,6 +918,7 @@ class Graph extends React.Component<{}, GraphState> {
           top: 0,
           left: 0,
         }}
+        data-test-id={`graph-wrapper-distance=${camera?.distance()}`}
```
atarashansky, line 921: I needed to added the camera's distance into the test attribute. Otherwise, I had no way of figuring out the zoom level since it's a virtual zoom on the coordinates, not a viewport zoom.

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,6 +43,18 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
   await page.mouse.up();
 }
 
+export async function scroll({testId, deltaY, coords}: {testId: string, deltaY: number, coords: number[]}) {
```
atarashansky, line 46: This function is for scrolling

client/__tests__/e2e/e2e.test.ts
```diff
@@ -444,6 +447,32 @@ describe("graph overlay", () => {
   });
 });
 
+test("zoom limit is 12x", async () => {
```
atarashansky, line 450: This test colors by and overlays centroid labels to be able to easily get a coordinate overlaying somewhere in the graph.

It then scrolls over that point and gets the test attribute from the graph wrapper to extract the new distance. The scroll was chosen to max out the zoom level, which should equal `scaleMax`.

client/src/components/graph/graph.tsx
seve: isn't this going to cause unnecessary re-renders?

seve (changes requested):

client/src/components/graph/graph.tsx
atarashansky: I figured if the camera is changing, the graph component would be re-rendered anyway. Let me double check if this is causing extra renders.

atarashansky: Doesn't seem to be causing extra renders when zooming.

client/src/components/graph/graph.tsx
```diff
@@ -126,6 +126,8 @@ class Graph extends React.Component<{}, GraphState> {
     return !shallowEqual(props.watchProps, prevProps.watchProps);
   }
 
+  private graphRef = React.createRef<HTMLDivElement>();
```
atarashansky, line 129: This is used to prevent scrolling on the graph component at mount time.

client/src/components/graph/graph.tsx
```diff
@@ -264,6 +266,11 @@ class Graph extends React.Component<{}, GraphState> {
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
   componentDidMount() {
     window.addEventListener("resize", this.handleResize);
+    if (this.graphRef.current) {
+      this.graphRef.current.addEventListener("wheel", this.disableScroll, {
```
atarashansky, line 270: prevent scrolling

client/src/components/graph/graph.tsx
atarashansky: After discussion with @tihuan, because the graph wrapper `div` element is re-rendering every time anything changes in the component anyway, including point position due to zoom, then adding dynamic `data-test-id` isn't causing extra renders. This is also confirmed by adding log statements to the component.

tihuan: Thanks for bringing this up, @seve! The `style={{...}}` is already re-generating a new POJO every time and causing re-rendering, so `data-test-id` changes is not making React render it more. It's already the most re-rendering React could do on this `<div>` 😄

seve (approved):

client/__tests__/e2e/e2e.test.ts
tihuan: feels like this comment is super helpful for understanding the test! Can we add it to the code please? Thank you!

client/src/components/graph/graph.tsx
tihuan: this is also a great context to have in the code 🙏

tihuan (approved):
LGTM! Just proposing to add some PR comments to code directly to beef up code context 🙏 Thanks so much!

Issue #564, opened 2023-02-15
Upgrade Explorer to Python 3.10


metakuni:
## Motivation

Language features availability, performance, security.

Secondarily, TileDB Mac M1/arm builds are available for Py 3.10+, which avoids a lengthy compilation process for Mac M1 users.

## Definition of Done

Explorer runs w/o issue on 3.10

## Tasks

- [ ] Docker containers updated
- [ ] Linters updated (to check against 3.10 compliance)
- [ ] Documentation updated
- [ ] Full system testing performed on rdev

PR #586, opened 2023-04-11
chore: update Docker base image to ubuntu 22

ebezzi:
Issue #564 

Update the Docker base image to Ubuntu 22. This implicitly brings the default Python version to 3.10.

### DIFF
--- a/hosted/Dockerfile
+++ b/hosted/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:focal
+FROM ubuntu:22.04
 
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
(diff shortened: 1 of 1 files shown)


hosted/Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV DEPLOYMENT_STAGE=$HAPPY_ENV
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y gettext nginx moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq npm git zip curl && \
+    apt-get install -y gettext nginx moreutils build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp llvm jq git zip curl && \
```
ebezzi, line 11: I think `npm` was wrong here. This installed a node version that was then overridden by the following install line. @tihuan can you verify if this makes sense?

tihuan: ooooh good call! Yeah that sounds right to me. And if the app compiles then LGTM lol

tihuan (approved):
LGTM!!

PR #633, opened 2023-07-06
fix: Allow Connection to Hubspot

ashin-czi:
Cannot make a request to Hubspot

image

modifying headers to match GE: https://github.com/chanzuckerberg/single-cell-data-portal/pull/4679/files#diff-e064de65d4610383e665a32e7e32ee1f520c466a6a63ecea8925ff8337f08215:~:text=frontend/next.config.js

### DIFF
--- a/server/ecs/app.py
+++ b/server/ecs/app.py
@@ -65,6 +65,7 @@ class WSGIServer(Server):
 
         csp = {
             "default-src": ["'self'", HUBSPOT_FORMS_URL, HUBSPOT_JS_URL],
+            "form-action": ["'self'", HUBSPOT_FORMS_URL],
             "connect-src": ["'self'", PLAUSIBLE_URL] + extra_connect_src,
             "script-src": ["'self'", "'unsafe-eval'", PLAUSIBLE_URL] + script_hashes,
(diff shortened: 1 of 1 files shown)


server/ecs/app.py
```diff
@@ -59,10 +59,15 @@ def _before_adding_routes(app, app_config):
...
 
+        HUBSPOT_JS_URL = "https://js.hsforms.net"
+
+        HUBSPOT_FORMS_URL = "https://forms.hsforms.com"
+
         csp = {
-            "default-src": ["'self'"],
-            "connect-src": ["'self'", PLAUSIBLE_URL] + extra_connect_src,
-            "script-src": ["'self'", "'unsafe-eval'", PLAUSIBLE_URL] + script_hashes,
+            "default-src": ["'self'", HUBSPOT_FORMS_URL],
```
tihuan, line 67: Seems like if we have the form URL in default src, the URL won't need to be in other directives 😄?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src

tihuan (approved):
LGTM! Just one non-blocking question. Thanks so much for resolving this, Andrew 🙏

PR #651, opened 2023-09-12
chore: update happy config task launch type

danieljhegeman:
Update happy config to work with current happy versions


Bento007 (approved):
Do we have explorer ec2 instance running that we can spin down after this?

danieljhegeman:
> Do we have explorer ec2 instance running that we can spin down after this?

Good question.

PR #653, opened 2023-09-13
feat: drop int64 fields from cxg_schema in the obs array

joyceyan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-explorer/644

the approach for solving this is to:
- allow int64 fields to be generated in the schema hints. this ensures that we can infer the type as int64 correctly
- filter out int64 fields when we're going through the `obs` array's schema to generate the dataset schema

a test was written by:
- generating `dataset_with_int64.cxg` using the following python notebook, and including some jank temporary changes to data portal to allow int64 in `type_conversion_utils.py`
```
import anndata
X = anndata.read_h5ad('local.h5ad')
import numpy as np
X.obs['int64_dtype'] = np.arange(X.shape[0], dtype='int64')
X.write_h5ad("modified.h5ad")
from backend.layers.processing.h5ad_data_file import H5ADDataFile
m = H5ADDataFile("modified.h5ad")
m.to_cxg("dataset-with-int64.cxg",100)
```
- these were the lines aded to `TypeConversionUtils._get_type_info_from_dtype` in data portal to allow generating a cxg dataset with int64 fields
```
    if dtype == np.dtype(np.int64):
        print("int64 dtype")
        return (np.int64, {"type": "int64"})
```
- then, i generated a fixture `schema_without_int64.json` which contains the expected schema after the int64 field is removed
- i confirmed that the int64 field was indeed added to this dataset because before i fixed the application code, the test was failing :)


server/dataset/cxg_dataset.py
```diff
@@ -381,6 +381,26 @@ def get_embedding_names(self):
             raise DatasetAccessError("cxg matrix missing embeddings")
         return embeddings
 
+    def _get_filtered_cxg_schema_metadata(self) -> list[Any]:
```
joyceyan, line 384: my understanding is that `obs.meta["cxg_schema"]` is just a json blob and we can filter out int64 fields here, but i'm not entirely sure what the structure of this metadata looks like

server/dataset/cxg_dataset.py
```diff
@@ -393,7 +427,13 @@ def _get_schema(self):
         annotations = {}
         for ax in ("obs", "var"):
             A = self.open_array(ax)
-            schema_hints = json.loads(A.meta["cxg_schema"]) if "cxg_schema" in A.meta else {}
+
+            if ax == "obs":
+                # Filter out int64 values in the obs cxg_schema metadata
+                schema_hints = self._get_filtered_cxg_schema_metadata()
```
joyceyan, line 433: not sure if it's a problem to open up the `obs` array twice, both in `A = self.open_array(ax)` and again in `self._get_filtered_cxg_schema_metadata()`

i would expect tiledb can handle concurrent reads without blowing up, but it could result in performance being less efficient.

server/dataset/cxg_dataset.py
```diff
@@ -343,14 +344,13 @@ def query_obs_array(self, term_name):
...
-        obs = self.open_array("obs")
-        meta = json.loads(obs.meta["cxg_schema"])
+        meta = self._get_filtered_cxg_schema_metadata()
         index_name = meta["index"]
         return index_name
 
     def get_obs_index(self):
         obs = self.open_array("obs")
-        meta = json.loads(obs.meta["cxg_schema"])
+        meta = self._get_filtered_cxg_schema_metadata()
```
joyceyan, line 353: not sure if it's a problem to open up the obs array twice, both in `obs = self.open_array("obs")` and again in `self._get_filtered_cxg_schema_metadata()`

i would expect tiledb can handle concurrent reads without blowing up, but it could result in performance being less efficient.

server/dataset/cxg_dataset.py
joyceyan, line 433: same issue here with opening up the `obs` array twice

atarashansky: https://github.com/chanzuckerberg/single-cell-explorer/blob/025beb186f5d2f718ec6cf7446032e3b59d6aa1c/server/dataset/cxg_dataset.py#L198

not sure what you're referring to, but if it's this function "open_array", then it's not opening multiple connections to TileDB. Just returning the same already opened connection from a hash table.

atarashansky (commented):
Looks good on first inspection - will wait for this to be tested before approving.

server/dataset/cxg_dataset.py
joyceyan: awesome thanks!

server/dataset/cxg_dataset.py
```diff
@@ -393,7 +422,13 @@ def _get_schema(self):
         annotations = {}
         for ax in ("obs", "var"):
             A = self.open_array(ax)
-            schema_hints = json.loads(A.meta["cxg_schema"]) if "cxg_schema" in A.meta else {}
+
+            if ax == "obs":
+                # Filter out int64 values in the obs cxg_schema metadata
+                schema_hints = self._get_filtered_cxg_schema_metadata() if "cxg_schema" in A.meta else {}
```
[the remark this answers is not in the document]
joyceyan: agreed - updated!

server/common/utils/type_conversion_utils.py
```diff
@@ -97,6 +97,9 @@ def _get_type_info_from_dtype(dtype) -> Union[Tuple[np.dtype, dict], None]:
     if dtype.kind == "O" and dtype.name != "category":
         return (np.dtype(str), {"type": "string"})
 
+    if allow_int64 and dtype.kind in ["i", "u"] and np.can_cast(dtype, np.int64):
```
atarashansky, line 100: This will never execute, because:

```
    if dtype.kind in ["i", "u"] and np.can_cast(dtype, np.int32):
        return (np.int32, {"type": "int32"})
```
will always execute before the allow_int64 block can execute.

Note that any integer dtype can be cast into int32 (including np.int64).

atarashansky (changes requested):

atarashansky (approved):
LGTM!!

PR #672, opened 2023-10-20
feat: add mypy to the server/common/utils directory

joyceyan:
## Reason for Change

- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-explorer/668

## Changes

- adds mypy config to `pyproject.toml` and uses the `server.common.utils` package as the guinea pig
- adds a make command, `make check`, which runs the pre-commit checks as well as `mypy --config-file pyproject.toml`
- ran the `run_mypy_and_annotate.py` script: https://github.com/chanzuckerberg/single-cell-ops-tools/blob/main/dev/run_mypy_and_annotate.py
- adds a pre-commit hook to run mypy
- note that since explorer uses python 3.8, running mypy found a few places where we use `list[str]` which requires python 3.9. i updated this use python 3.8 friendly code instead, `List[str]`, importing `List` from `typing`
  - i'm not sure what the implications of upgrading to python 3.9+ would be, but it does seem a bit odd that there's python 3.9+ code in a repo that supposedly requires python 3.8 🤔 

## Testing

- ran the script, and verified `make check` can run with no issues

## Notes for Reviewer


joyceyan:
commenting so this doesn't get removed

pyproject.toml
```diff
@@ -28,3 +28,14 @@ select = ["E", "F", "B", "SIM", "C4", "I"]
 ignore = ["E501", "E402", "C408", ]
 line-length = 120
 target-version = "py38"
+
+[tool.mypy]
+# here is where we tell mypy to only check the scripts package. note that we're using
+# packages = ["scripts"] instead of modules = ["scripts"] because packages will recursive
```
atolopko-czi, line 34:
```suggestion
# packages = ["scripts"] instead of modules = ["scripts"] because packages will recursively
```

atolopko-czi, line 34:
```suggestion
# packages = ["server.common.utils"] instead of modules = ["server.common.utils"] because packages will recursive
```

atolopko-czi (approved):
Suggested a couple comment corrections, but otherwise LGTM

pyproject.toml
joyceyan: ah good catch, thank you!

Issue #679, opened 2023-11-22
[BUG] Improper display of genes with non-zero expression in single-cell-explorer quick gene search


MaximilianLombardo:
**Describe the bug**
A user [identified](https://github.com/chanzuckerberg/cellxgene-census/issues/862) that plotting gene expression via the quick gene search not functional in all of the explorer instances for this collection: [MSK SPECTRUM – Ovarian cancer mutational processes drive site-specific immune evasion](https://cellxgene.cziscience.com/collections/4796c91c-9d8f-4692-be43-347b1727f9d8).

When searching for a gene in the quick gene search, the gene appears and can be added to the UI. If the gene has zero expression, then the gene should not appear in quick gene search. Observed behavior deviates from expected behavior because for a gene that has non-zero expression for any cells ([ex: dendritic cells](https://cellxgene.cziscience.com/e/97d9238c-1a39-4873-b0bb-963ec2d788e6.cxg/)), the expected histogram of gene expression values (distribution of expression values across all cells in the dataset) should appear. The observed behavior is pictured below

<img width="365" alt="image" src="https://github.com/chanzuckerberg/single-cell-explorer/assets/25663501/661dc161-a18a-4a89-ad4a-dfda32a9fdbc">

A cursory check of other collections in the data portal seem to indicate that this is the only collection affected.

**To Reproduce**
1. open [MSK Spectrum Collection](https://cellxgene.cziscience.com/collections/4796c91c-9d8f-4692-be43-347b1727f9d8)
2. Open the dendritic cells explorer instance
3. Plot a gene (I've tried MALAT1, ATM, ACTB which all generate the unintended behavior)
4. Observe that instead of histogram, you get an ":0"

**Expected behavior**
Described in detail above - histogram showing the distribution of expression values in cells where the gene was detected.

**Screenshots**
Error screenshots above. I also downloaded the object to manually confirm that there were non-zero expression values in `adata.X` for an example gene ['MALAT1'] that manifests the error in the hosted explorer instance. See screenshot below:

![image](https://github.com/chanzuckerberg/single-cell-explorer/assets/25663501/ef52de57-006c-4481-b7a4-c16118756a67)

**Version (please complete the following information):**
 - Browser [chrome]:
 - Version [Version 119.0.6045.159 (Official Build) (x86_64)]
 - MacOS Sonoma 14.1.1

**Additional context**
[...]

PR #684, opened 2023-12-12
fix(explorer): filter out where feature_is_filtered is true

kaloster:
#679 

Added logic to exclude genes where `feature_is_filtered:True` from the Quick Gene Search

### DIFF
--- a/client/src/components/geneExpression/quickGene.tsx
+++ b/client/src/components/geneExpression/quickGene.tsx
@@ -61,6 +61,8 @@ function QuickGene() {
           const df: Dataframe = await annoMatrix.fetch("var", varIndex);
           let dfIds: Dataframe;
           const geneIdCol = "feature_id";
+          const isFilteredCol = "feature_is_filtered";
+          const isFiltered = await annoMatrix.fetch("var", isFilteredCol);
 
           // if feature id column is available in var
           if (annoMatrix.getMatrixColumns("var").includes(geneIdCol)) {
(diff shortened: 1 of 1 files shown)


client/src/components/geneExpression/quickGene.tsx
```diff
@@ -69,7 +71,18 @@ function QuickGene() {
...
 
           setStatus("success");
-          setGeneNames(df.col(varIndex).asArray() as DataframeValue[]);
+
+          setGeneNames(
+            df
+              .col(varIndex)
+              .asArray()
+              .filter((_, index) => {
+                const isFilteredValue = isFiltered.col(isFilteredCol).asArray()[
```
atarashansky, line 80: i doubt this materializes the full array in memory on every iteration of the loop as these chained functions are probably accessors, but just to be safe, can we save the isFiltered array outside the loop to a constant?

client/src/components/geneExpression/quickGene.tsx
```diff
@@ -197,7 +209,7 @@ function QuickGene() {
               inputProps={{
                 // @ts-expect-error ts-migrate(2322) FIXME: Type '{ "data-testid": string; placeholder: string... Remove this comment to see the full error message
                 "data-testid": "gene-search",
-                placeholder: "Quick Gene Search",
+                placeholder: "Quick Gene Search!",
```
atarashansky, line 212: Is the exclamation point part of the ticket?

atarashansky (changes requested):

atarashansky (approved):

PR #704, opened 2024-01-17
chore(explorer): enable safari fix browser page

kaloster:
[Zenhub Ticket](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/gh/chanzuckerberg/single-cell-data-portal/6429)

Updates:
- enabled Safari on browserlist, and created new regex as per [here](https://github.com/chanzuckerberg/single-cell-explorer/pull/159)
- fixed broken images on Unsupported Browser page, added Safari to supported browsers
- updated webpack's `file-loader` to `asset/resource` type
Screenshot 2024-01-17 at 10 35 43 AM


client/index.html
```diff
@@ -59,7 +59,7 @@
     <div id="root"></div>
     <script
       async
-      src="static/obsoleteBrowsers.js"
+      src="/static/obsoleteBrowsers.js"
```
tihuan, line 62: I believe the `/` prefix was only for local testing, so might need to reverse this back. Thank you!

client/configuration/webpack/obsoleteHTMLTemplate.html
```diff
@@ -38,27 +35,35 @@
       <a
         href="https://www.google.com/chrome/?hl=en%22"
         aria-label="Download Google Chrome"
+        target="_blank"
       >
-        <img
-          src="https://cellxgene.cziscience.com/s3/cellxgene/static/images/chrome.png"
-          style="width: 80px; height: 80px"
-        />
+        <img src="/images/chrome.png" style="width: 80px; height: 80px" />
```
tihuan, line 40: We'll need to see in `dev` if we need to remove the `/` prefix in `src`, since I think it was just for local testing and might break in deployed environment!

client/src/index.tsx
```diff
@@ -9,6 +9,12 @@ import "./index.css";
 import App from "./components/app";
 import store from "./reducers";
 
+import "./assets/img/safari.png";
```
tihuan, line 12: can we add a comment to indicate that these are imported for Webpack `asset/resource` module to bundle in order to be used in obsolete browser template 🙏 ?

Thank you!

tihuan (approved):
Wow even the logo looks better! Thanks so much, @kaloster 😄 🙏 

Just a few tiny nits, otherwise LGTM!

PR #709, opened 2024-01-18
feat(playwright): migrate tests and e2e utils

seve:
This PR ports over the entire test suite over to playwright and playwright test runner. This includes a large amount of test API changes, as well as some modifications into how the tests themselves are run and how assertions are made.  I'm not sure if unit tests are currently run in our CI/CD pipeline, but I made edits to those files accordingly.  To run locally, run your fe server and pass the url to the e2e command like so:
`CXG_URL_BASE="http://localhost:3000" npm run e2e`

There are still a few failing tests on the GH workflow:
* "categories and values from dataset appear and properly truncate if applicable"
* "continuous data appears"
* "select cells via lasso"

Unfortunately these tests are different from those failing locally, so debugging is a bit difficult.

----
Note: For some reason, pre-commit didn't catch a lot of tslint errors, all of which weren't due to edits.  If someone could aid me in fixing/removing these lint issues, that would be fantastic.


client/__tests__/common/constants.ts
```diff
@@ -1,25 +1,20 @@
...
-};
+export const APP_URL_BASE =
+  process.env.CXG_URL_BASE || `http://localhost:${ENV_DEFAULT.CXG_CLIENT_PORT}`;
 
-export const TEST_URL = TEST_ENV_TO_TEST_URL[TEST_ENV];
+const DEFAULT_BASE_PATH = "d";
+export const TEST_URL = APP_URL_BASE.includes("localhost")
+  ? [APP_URL_BASE, DEFAULT_BASE_PATH, DATASET].join("/")
+  : APP_URL_BASE;
+export const PAGE_URL_TRUNCATE = [
```
tihuan, line 13: let's use regular variableName instead of ALL_CAPS here, since the value is not hard coded like a real constant!

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
+  deltaY,
+  coords,
+  page,
+}: {
+  testId: string;
+  deltaY: number;
+  coords: number[];
+  page: Page;
+}): Promise<void> {
+  const layout = await page.getByTestId(testId);
```
tihuan, line 51: no need to await here, since Playwright locator is designed to check the DOM at the time of user action.

E.g.,

```ts
const button = page.getByTestId("test-button");
await button.click() // this we will await!
```

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
-
-          return [cat, count];
-        })
-      )
+  const categoryRows = await page
+    .getByTestId(`category-${category}`)
+    .getByTestId("categorical-row")
+    .all();
+
+  return Object.fromEntries(
```
tihuan, line 103: Discussion: would `Array.from()` work here? Feels more intuitive, given that we're working with array

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
 
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-export async function getCellSetCount(num: any) {
-  await clickOn(`cellset-button-${num}`);
-  return getOneElementInnerText(`[data-testid='cellset-count-${num}']`);
+export async function getCellSetCount(
+  num: number,
+  page: Page
+): Promise<string> {
+  await page.getByTestId(`cellset-button-${num}`).click({ force: true });
```
tihuan, line 125: just curious, why is `force` needed? Is there underlying issue we should fix?

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-export async function resetCategory(category: any) {
+export async function resetCategory(
+  category: string,
+  page: Page
+): Promise<void> {
   const checkboxId = `${category}:category-select`;
-  await waitByID(checkboxId);
-  const checkedPseudoclass = await page.$eval(
+  await page.getByTestId(checkboxId);
```
tihuan, line 134: this doesn't do anything, since locator doesn't return a promise

If we wanna wait for the element to be visible, maybe:

```ts
await page.getByTestId(checkboxId).waitFor({ state: "visible" });
```

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
+  const checkedPseudoClass = await page.$eval(
     `[data-testid='${checkboxId}']`,
     (el) => el.matches(":checked")
   );
-  if (!checkedPseudoclass) await clickOn(checkboxId);
+  if (!checkedPseudoClass)
+    await page.getByTestId(checkboxId).click({ force: true });
```
tihuan, line 140: if `force` is needed, let's add a comment to explain why, so we can revisit the need to `force` if the condition changes in the future!

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
 
-  const categoryRow = await waitByID(`${category}:category-expand`);
```
tihuan, line 151: Should we update `waitByID` utility function to be the following?

```ts
function waitByID({ testId, page }) {
  return page.getByTestId(testId).waitFor({ state: "visible" });
}
```

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
-  const el = await waitByID(testId);
-  // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
-  const size = await el.boxModel();
+  testId: string,
+  xAsPercent: number,
+  yAsPercent: number,
+  page: Page
+): Promise<Coordinate> {
+  const el = await page.getByTestId(testId);
+  const size = await el.boundingBox();
```
tihuan, line 159:
```suggestion
  const size = await page.getByTestId(testId).boundingBox();
```

client/__tests__/e2e/cellxgeneActions.ts
seve: check to see what layout is and why we need to wait and what we should be waiting for (visible?)

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
+  page: Page
+): Promise<void> {
   const checkboxId = `${category}:category-select`;
-  await waitByID(checkboxId);
-  const checkedPseudoclass = await page.$eval(
+  await page.getByTestId(checkboxId);
+  const checkedPseudoClass = await page.$eval(
     `[data-testid='${checkboxId}']`,
     (el) => el.matches(":checked")
   );
```
seve, line 138: remove eval in favor playwright api getAttribute

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,262 +37,256 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
+export async function assertColorLegendLabel(
+  label: string,
+  page: Page
+): Promise<void> {
+  const handle = await page.getByTestId("continuous_legend_color_by_label");
 
-  // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
   const result = await handle.evaluate((node) =>
     node.getAttribute("aria-label")
   );
```
seve, line 292: Do not eval instead use get attribute

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -327,158 +316,152 @@ export async function addGeneToSetAndExpand(
...
   const submitButton = `${genesetName}:submit-gene`;
-  await clickOnUntil(`${genesetName}:add-new-gene-to-geneset`, async () => {
-    await expect(page).toMatchElement(getTestId("add-genes"));
-  });
-  assert(await isElementPresent(getTestId("add-genes"), {}));
-  await typeInto("add-genes", geneToAddToSet);
-  await clickOn(submitButton);
+  await page.getByTestId(`${genesetName}:add-new-gene-to-geneset`).click();
+  await page.getByTestId("add-genes-to-geneset-dialog").click();
+  await page.getByTestId("add-genes-to-geneset-dialog").fill(geneToAddToSet);
```
seve, line 398: assert that text input is filled and/or use `pressSequentially`

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -327,158 +316,152 @@ export async function addGeneToSetAndExpand(
...
-  // @ts-expect-error ts-migrate(2554) FIXME: Expected 2 arguments, but got 1.
-  const result = await isElementPresent(
-    getTestId(`${genesetName}:geneset-name`)
-  );
-  await expect(result).toBe(false);
+export async function assertGenesetDoesNotExist(
+  genesetName: string,
+  page: Page
+): Promise<void> {
+  await expect(page.getByTestId(`${genesetName}:geneset-name`)).toBeFalsy();
```
seve, line 370: tryUntil?

client/__tests__/e2e/cellxgeneActions.ts
seve: try `trial` option on click to do try run and check if dialog was filled

client/__tests__/e2e/e2e.test.ts
```diff
@@ -98,435 +94,464 @@ const genesetToCheckForDescription = "fourth_gene_set";
...
   });
 
-  test("continuous data appears", async () => {
-    await goToPage(pageUrl);
-
+  test("continuous data appears", async ({ page }) => {
+    await goToPage(page);
     for (const label of Object.keys(data.continuous)) {
-      await waitByID(`histogram-${label}`);
+      expect(await page.getByTestId(`histogram-${label}-plot`)).not.toHaveCount(
```
seve, line 184: wrap in try until

client/__tests__/e2e/e2e.test.ts
```diff
@@ -98,435 +94,464 @@ const genesetToCheckForDescription = "fourth_gene_set";
...
 
     // Toggle colorby for each category and check to see if labels are generated
     for (let i = 0, { length } = labels; i < length; i += 1) {
       const label = labels[i];
       // first label is already enabled
-      if (i !== 0) await clickOn(`colorby-${label}`);
-      const generatedLabels = await getAllByClass("centroid-label");
+      if (i !== 0) await page.getByTestId(`colorby-${label}`).click();
+      const generatedLabels = await page.getByTestId("centroid-label").all();
+      console.log(generatedLabels);
```
seve, line 396:
```suggestion
```

PR #701, opened 2024-01-09
Install and setup playwright, successfully run example test

seve:
This is a breaking change to the current test suite, the existing tests still need to be migrated onto playwright. All e2e GHA will be failing as a result.

To test:

npm install and run the following command

```
npx playwright test __tests__/example.spec.ts
```

PR #714, opened 2024-01-23
test(playwright): Migrate jest + puppeteer to playwright

seve:
Migrates explorer's e2e + frontend unit test suites from jest and puppeteer to playwright.

Some tests have been disabled as the work required to get them to pass (they have been migrated in the below PR's) began to get too large.
 * "categories and values from dataset appear and properly truncate if applicable"
 * The following are only failing on GHA, most likely due to a race condition with undo/redo (#715)
   * "create a new geneset and undo/redo"
   * "edit geneset name and undo/redo"
   * "delete a geneset and undo/redo"
   * "delete gene from geneset and undo/redo"

See: 
#709
#701


.github/workflows/push_tests.yml
```diff
@@ -98,16 +96,32 @@ jobs:
...
+          path: /Users/runner/work/single-cell-explorer/single-cell-explorer/client/playwright-report
+          retention-days: 14
+
+      - name: Upload blob report to GitHub Actions Artifacts
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          name: all-blob-reports
+          path: /Users/runner/work/single-cell-explorer/single-cell-explorer/client/blob-report
+          retention-days: 1
```
tihuan, line 127: thanks so much for going the extra mile to upload the artifact! 🔥 Super helpful for debugging 🙏

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,282 +38,269 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
+
+export async function waitUntilNoSkeletonDetected(page: Page): Promise<void> {
+  await tryUntil(
+    async () => {
+      const skeleton = await page
+        .locator(`.${BLUEPRINT_SKELETON_CLASS_NAME}`)
+        .all();
+      expect(skeleton).toHaveLength(0);
+    },
+    { page, timeoutMs: 10000 }
```
tihuan, line 79: I think we can do this now lol?
```suggestion
    { page, timeoutMs: 10_000 }
```

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,282 +38,269 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
-  const checkedPseudoclass = await page.$eval(
-    `[data-testid='${checkboxId}']`,
-    (el) => el.matches(":checked")
-  );
-  if (!checkedPseudoclass) await clickOn(checkboxId);
+export async function resetCategory(
+  category: string,
+  page: Page
+): Promise<void> {
+  const checkbox = await page.getByTestId(`${category}:category-select`);
```
tihuan, line 138: no need to await!

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -43,282 +38,269 @@ export async function drag(testId: any, start: any, end: any, lasso = false) {
...
 
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-export async function getCellSetCount(num: any) {
-  await clickOn(`cellset-button-${num}`);
-  return getOneElementInnerText(`[data-testid='cellset-count-${num}']`);
+export async function getCellSetCount(
+  num: number,
+  page: Page
+): Promise<string> {
+  await page.getByTestId(`cellset-button-${num}`).click({ force: true });
```
tihuan, line 130: I think it might be work adding a comment to explain why `{ force: true }` is used here. I know you verbally told Ronen and I why, but Alec and the rest of the team would still love the reminder 😄 Thank you!

here and throughout

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -327,158 +309,198 @@ export async function addGeneToSetAndExpand(
...
-  const result = await isElementPresent(getTestId(`${geneSymbol}:gene-label`));
+export async function assertGeneDoesNotExist(
+  geneSymbol: string,
+  page: Page
+): Promise<void> {
+  await tryUntil(
+    async () => {
+      const geneLabel = await page.getByTestId(`${geneSymbol}:gene-label`);
+      await geneLabel.waitFor({ state: "hidden" });
+      await expect(geneLabel).toBeHidden();
```
tihuan, line 461: do we need this if the previous step is already checking it?

client/__tests__/e2e/e2e.test.ts
```diff
@@ -85,448 +78,477 @@ const setToRemoveFrom = "empty_this_geneset";
...
+  test("color by", async ({ page }) => {
+    await goToPage(page);
     const allLabels = [
       ...Object.keys(data.categorical),
       ...Object.keys(data.continuous),
     ];
 
     for (const label of allLabels) {
-      await clickOn(`colorby-${label}`);
+      await page.getByTestId(`colorby-${label}`).click();
```
tihuan, line 358: Maybe worth a ticket to capture this, but this and the original test don't seem to assert after clicking the element it does color the umap 😆

client/__tests__/e2e/e2e.test.ts
```diff
@@ -85,448 +78,477 @@ const setToRemoveFrom = "empty_this_geneset";
...
       "layout-graph",
-      data.pan["coordinates-as-percent"]
+      data.pan["coordinates-as-percent"],
+      page
     );
 
-    await drag("layout-graph", panCoords.start, panCoords.end, false);
+    await drag("layout-graph", panCoords.start, panCoords.end, page);
 
     await page.evaluate("window.scrollBy(0, 1000);");
```
tihuan, line 373: same for this one - not sure how this test can detect when pan/zoom breaks, since it just performs actions but not actually asserting the UI actually changes somehow

Maybe a ticket to actually assert it if possible!

If we can't use element attributes to detect changes, we also now have early access to Chromatics e2e test for visual testing 😄 !

If it'll make these types of testing easier, we should create a ticket to set it up!

https://www.chromatic.com/blog/e2e-visual-tests-sneak-peek/

client/__tests__/e2e/e2e.test.ts
```diff
@@ -85,448 +78,477 @@ const setToRemoveFrom = "empty_this_geneset";
...
+          `centroid-label`,
+          categoryValue,
+          page
+        );
+
+        expect(terminalCoordinates[0] - initialCoordinates[0]).toBeCloseTo(
+          panCoords.end.x - panCoords.start.x
+        );
+        expect(terminalCoordinates[1] - initialCoordinates[1]).toBeCloseTo(
+          panCoords.end.y - panCoords.start.y
```
tihuan, line 440: oh this asserts the dragging. Maybe we can do something similar for the tests above in the new tickets!

client/__tests__/e2e/playwright.global.setup.ts
```diff
@@ -0,0 +1,45 @@
+import { Page } from "@playwright/test";
+import { DATASET_METADATA_RESPONSE } from "../__mocks__/apiMock";
+
+const setup = async ({ page }: { page: Page }) => {
+  await page.route("**/*/dataset-metadata", (route, request) => {
```
tihuan, line 5: let's add some comments to provide context why we're mocking these calls!

client/__tests__/e2e/test_config.yaml
```diff
@@ -3,7 +3,7 @@ server:
     force_https: true
     port: 5005
     debug: true
-    verbose: true
+    verbose: false
```
tihuan, line 6: for my own education: why is this `false` now?

client/__tests__/util/helpers.ts
```diff
@@ -0,0 +1,273 @@
+import { ElementHandle, expect, Locator, Page } from "@playwright/test";
+
+import { ERROR_NO_TEST_ID_OR_LOCATOR } from "../common/constants";
+import { waitUntilNoSkeletonDetected } from "../e2e/cellxgeneActions";
+
+export const TIMEOUT_MS = 3 * 1000;
+export const WAIT_FOR_TIMEOUT_MS = 3 * 1000;
+
+// (seve): We use TEST_ENV to describe the environment that playwright is running against. Sometimes the FE tests are run against a local instance of the app which points at a deployed instance of the backend.
```
tihuan, line 9: wait I don't see `TEST_ENV` variable being used anywhere. Maybe I'm missing something here lol

client/src/components/framework/toasters.ts
```diff
@@ -1,40 +1,43 @@
...
-  position: Position.TOP,
-  maxToasts: 4,
-});
+if (typeof document !== "undefined") {
+  ToastTopCenter = Toaster.create({
+    className: "recipe-toaster",
+    position: Position.TOP,
+    maxToasts: 4,
+  });
+}
```
tihuan, line 16: in React 18 + Blueprint 5, we'll need to update how this is instantiated here. But we will have example in DP to follow!

client/src/components/geneExpression/menus/editGenesetNameDialogue.tsx
```diff
@@ -138,7 +138,7 @@ class RenameGeneset extends React.PureComponent<{}, State> {
               // @ts-expect-error ts-migrate(2322) FIXME: Type '{ label: any; onChange: (e: any) => void; in... Remove this comment to see the full error message
               label={newGenesetDescription}
               onChange={this.handleChangeDescription}
-              inputProps={{ "data-testid": "change geneset description" }}
+              inputProps={{ "data-testid": "change-geneset-description" }}
```
tihuan, line 141: thanks for adding the kebab-case!

tihuan (commented):
Looks super good! Thanks SO much for the extraordinary work to get Explorer to be on Playwright!! So exciting 👏 🥇 👏 !!

client/__tests__/e2e/cellxgeneActions.ts
seve:
```suggestion
  const checkbox = page.getByTestId(`${category}:category-select`);
```

client/__tests__/e2e/test_config.yaml
seve: It was an attempt to clean up the GHA logs for smoke tests.  The info level logs from the cellxgene server aren't necessary for the smoke tests, and this verbose config attribute _should_ be connected to that feature. I think this aspect of the config is broken though.

client/__tests__/e2e/cellxgeneActions.ts
seve: I'll add a comment in at the top of the file

client/__tests__/e2e/cellxgeneActions.ts
seve: technically no, but it felt wrong to have a function `assertGeneDoesNotExist` without an actual assertion

client/__tests__/util/helpers.ts
seve: accidentally left in from dp copy

client/__tests__/e2e/puppeteer.setup.ts
```diff
@@ -1,107 +0,0 @@
...
-        }
-        const errorMsgText = await Promise.all(
-          // TODO can we do this without internal properties?
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-          msg.args().map((arg: any) => arg._remoteObject.description)
-        );
-        throw new Error(`Console error: ${errorMsgText}`);
-      }
-    }
-  });
```
seve, line 106: old error handling

client/__tests__/e2e/e2e.test.ts
seve: there aren't any elements to track for the above 😢

client/__tests__/e2e/cellxgeneActions.ts
tihuan: Got it! Sounds good 👍

client/__tests__/e2e/test_config.yaml
tihuan: ah so changing this doesn't produce less logs? If so should we just revert it back to `true` for now until we figure out whether we should even delete the config here?

client/__tests__/e2e/puppeteer.setup.ts
tihuan: Sorry I forgot - did you create a ticket to capture the work to reintroduce this 😄 ? Thanks!

client/__tests__/e2e/e2e.test.ts
tihuan: Oy got it. Chromatic it is 🎉 !

client/__tests__/e2e/test_config.yaml
seve: Config is definitely needed.  I can delete this line if we want since it defaults to `false`, but I think `false` is the correct value

client/__tests__/e2e/e2e.test.ts
seve: #716

client/__tests__/e2e/e2e.test.ts
seve: #716

client/__tests__/e2e/puppeteer.setup.ts
seve: https://github.com/chanzuckerberg/single-cell-explorer/issues/719

client/__tests__/e2e/test_config.yaml
tihuan: Oh I meant to restore the value back to `true`, since that’s what it was before right?

And since changing it to `false` is not reducing the logs, it feels like we don’t know what impact changing it to `false` actually does, so I was thinking to be safe, maybe just revert it back to `true`, so we don’t accidentally break something we don’t know

client/__tests__/e2e/puppeteer.setup.ts
tihuan: Thanks! Linked to the visium epic 👍

client/__tests__/e2e/test_config.yaml
seve:
```suggestion
    verbose: true
```

tihuan (approved):
Fantastic work, Seve! Really appreciate the quick iterations to get this PR through 👏 !!

PR #788, opened 2024-02-23
chore(ts): Type the graph, lasso, and some rendering code

seve:
Broke out my typing into a separate PR. Found a couple of existing bugs, but I have little doubt that I introduced something. These are a lot of changes to not often touched code, so please take your time reviewing!


client/src/components/graph/setupSVGandBrush.ts
seve, line 1: Separated the two different setups into two functions. Out of scope for the PR, but I'm pretty sure brush has been dead for a while now, we can safely delete all brush code

client/src/components/graph/graph.tsx
```diff
@@ -306,62 +308,58 @@ class Graph extends React.Component<{}, GraphState> {
...
-      // @ts-expect-error ts-migrate(2339) FIXME: Property 'toolSVG' does not exist on type '{}'.
       stateChanges.toolSVG
     ) {
       const { tool, container } = this.state;
       this.selectionToolUpdate(
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'tool' does not exist on type '{}'.
-        stateChanges.tool ? stateChanges.tool : tool,
-        // @ts-expect-error ts-migrate(2339) FIXME: Property 'container' does not exist on type '{}'.
-        stateChanges.container ? stateChanges.container : container
+        stateChanges.tool ? stateChanges.tool : tool!,
```
seve, line 317: There's a lot of non-null assertions because state is initiated as null, but quickly setup up after mounting

client/src/components/graph/graph.tsx
```diff
@@ -557,32 +523,37 @@ class Graph extends React.Component<{}, GraphState> {
...
     } else {
       handleStart = this.handleLassoStart.bind(this);
       handleEnd = this.handleLassoEnd.bind(this);
       handleCancel = this.handleLassoCancel.bind(this);
+      ret = setupLasso({
+        selectionToolType: selectionTool,
+        handleStartAction: handleStart,
+        handleEndAction: handleEnd,
+        handleCancelAction: handleCancel,
+      });
```
seve, line 549: Break tool creation into separate functions

client/src/components/graph/graph.tsx
```diff
@@ -817,23 +779,24 @@ class Graph extends React.Component<{}, GraphState> {
       needToRenderCanvas = true;
     }
     if (positions !== prevAsyncProps?.positions) {
-      pointBuffer({ data: positions, dimension: 2 });
+      // @ts-expect-error (seve): need to look into arg mismatch
+      pointBuffer!({ data: positions, dimension: 2 });
```
seve, line 783: Really not sure how to change these functions

kaloster (approved):
This looks good to me. Thank you @seve for typing this and cleaning up the codebase, it actually makes a huge difference.

What should be the merge strategy for this? Should we just merge yours main then update mine with main and work through any conflicts?

PR #792, opened 2024-02-26
feat: fix explorer API for handling subpaths in S3

atarashansky:
Fixed some bugs that were only visible when testing on a deployed environment. This has been tested with (way too many) pirated deployments to `dev` and all is working.

### DIFF
--- a/.happy/config.json
+++ b/.happy/config.json
@@ -1,49 +1,51 @@
 {
-    "config_version": "v1",
-    "terraform_version": "0.13.5",
-    "default_env": "rdev",
-    "app": "explorer",
-    "default_compose_env_file": ".env.ecr",
-    "services": [
-        "explorer"
-    ],
-    "environments": {
-        "rdev": {
-            "aws_profile": "single-cell-dev",
-            "secret_arn": "happy/env-explorer-dev-config",
-            "terraform_directory": ".happy/terraform/envs/rdev",
-            "log_group_prefix": "/explorer/rdev"
-        },
-        "dev": {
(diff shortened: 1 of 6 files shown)


.happy/config.json
atarashansky, line 1: Linting and added task_launch_type

client/src/globals.ts
atarashansky, line 1: Updating the REGEX to assume a 1-character dataroot so that we can get all paths leading from the dataroot up to /api. Otherwise, it is very difficult to generalize a regex that can take an aribtrary number of subpaths (e.g. `cellguide-cxgs/tissues/homo_sapiens/etc/api` without including the `/cellxgene/` API URL prefix.

server/app/app.py
```diff
@@ -204,9 +204,11 @@ def __init__(self, app_config: AppConfig):
                 methods=["GET"],
             )
             self.app.add_url_rule(
-                f"/{url_dataroot}/<path:dataset>.cxg/",
+                f"/{url_dataroot}/{CELLGUIDE_CXG_KEY_NAME}/<path:dataset>.cxg/",
```
atarashansky, line 207: This explicitly ensures that we only capture URLs that for this rule that are prefixed by "cellguide-cxgs/". <path:dataset> was interacting poorly with <string:dataset> because one subsumes the other.

.happy/config.json
atarashansky, line 1: Linted and added task_launch_type: ec2 which is now required for latest happy versions

prathapsridharan (approved):
lgtm

PR #796, opened 2024-02-28
feat(img): capture graph as a image for testing and user download

seve:
Adds functionality to enable capturing the webgl canvas as an image. This can now be used for testing as well as user download.

### DIFF
--- a/client/configuration/eslint/eslint.js
+++ b/client/configuration/eslint/eslint.js
@@ -106,22 +106,6 @@ module.exports = {
         "@typescript-eslint/no-var-requires": "off",
       },
     },
-    {
-      files: ["**/*.test.ts"],
-      env: {
-        jest: true, // now **/*.test.ts files' env has both es6 *and* jest
-      },
-      // Can't extend in overrides: https://github.com/eslint/eslint/issues/8813
-      // "extends": ["plugin:jest/recommended"]
-      plugins: ["jest"],
-      rules: {
-        "jest/no-disabled-tests": "warn",
-        "jest/no-focused-tests": "error",
-        "jest/no-identical-title": "error",
-        "jest/prefer-to-have-length": "warn",
-        "jest/valid-expect": "error",
-      },
-    },
   ],
 };
 /* eslint-enable @blueprintjs/classes-constants -- we don't import blueprint here  */
--- a/client/src/components/graph/graph.tsx
+++ b/client/src/components/graph/graph.tsx
@@ -1,5 +1,7 @@
 import React, { MouseEvent, MouseEventHandler } from "react";
 import * as d3 from "d3";
+import { toPng } from "html-to-image";
+
 import { connect, shallowEqual } from "react-redux";
 import { ReadonlyMat3, mat3, vec2 } from "gl-matrix";
 import _regl, { DrawCommand, Regl } from "regl";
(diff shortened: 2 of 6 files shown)


client/configuration/eslint/eslint.js
seve, line 1: we no longer use jest nor its eslint plugin

client/src/components/graph/graph.tsx
```diff
@@ -793,6 +810,9 @@ class Graph extends React.Component<GraphProps, GraphState> {
       flagBuffer!({ data: flags, dimension: 1 });
       needToRenderCanvas = true;
     }
+    if (screenCap) {
+      needToRenderCanvas = true;
+    }
```
seve, line 815: I believe it is safe to not check if screencap is differeing from the last call of updateReglAndRender, since this is just an on off state. I put a log in this function and played around and it affirmed.

client/src/components/graph/graph.tsx
```diff
@@ -856,6 +876,38 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+    if (screenCap && regl) {
+      const graph = regl._gl.canvas;
+      const imageURI = await toPng(graph as HTMLCanvasElement, {
+        backgroundColor: "white",
+        height,
+        width,
+      });
+      this.setState({ testImageSrc: imageURI });
+      try {
+        // TODO: DOWNLOAD IMAGE
```
seve, line 889: This can be used to download the image in the future.

client/src/components/graph/graph.tsx
```diff
@@ -930,6 +984,22 @@ class Graph extends React.Component<GraphProps, GraphState> {
           onDoubleClick={this.handleCanvasEvent}
           onWheel={this.handleCanvasEvent}
         />
+        {testImageSrc && (
```
seve, line 987: Checks if the imageSrc is not null and then populates an `` with the string

client/src/components/graph/graph.tsx
```diff
@@ -960,8 +1031,11 @@ class Graph extends React.Component<GraphProps, GraphState> {
             )}
           </Async.Rejected>
           <Async.Fulfilled>
-            {(asyncProps) => {
-              if (regl && !shallowEqual(asyncProps, this.cachedAsyncProps)) {
+            {(asyncProps: GraphAsyncProps) => {
+              if (
+                regl &&
+                (!shallowEqual(asyncProps, this.cachedAsyncProps) || screenCap)
```
seve, line 1037: simple addition to always update regl if screenCap is true

client/src/components/menubar/index.tsx
```diff
@@ -350,6 +354,24 @@ class MenuBar extends React.PureComponent<{}, State> {
...
+                type="button"
+                icon={IconNames.TORCH}
+                style={{
+                  cursor: "pointer",
+                }}
+                loading={screenCap}
+                onClick={() => dispatch({ type: "graph: screencap start" })}
+              />
+            </Tooltip>
+          )}
```
seve, line 374: Feature flag enabled button that will be enabled via playwright or manual QA

client/src/components/graph/graph.tsx
```diff
@@ -98,7 +100,15 @@ type GraphState = {
   };
   modelTF: ReadonlyMat3;
   modelInvTF: ReadonlyMat3;
+  testImageSrc: string | null;
```
tihuan, line 103: maybe worth adding a comment to explain the context on why we need this prop for future reference? Something along the line that "this is used for Chromatic visual test` or something better!

tihuan (approved):
This is super cool! Thanks so much for adding this, @seve 🙏 👏 !

Please wait for @atarashansky 's review, since both of you know Explorer better than I do 😆

PR #807, opened 2024-03-04
fix: fix regex API prefix substitution for dataset selection

atarashansky:
- Fixes an issue with the new REGEX not properly updating the API prefix when selecting a new dataset
 - Fixes an issue with trailing slashes when trying to access a CXG on localhost:3000 (this broke local e2e tests).

The "REGEX_PATHNAME_FOR_DATASET_SELECTOR" is what the original regex was prior to the changes made for enabling access to CXGs in S3 subfolders

Tested in a pirated `dev` deployment since we do not have test infrastructure for testing dataset selection.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -46,7 +46,7 @@
     <script type="text/javascript">
       window.CELLXGENE = {};
       window.CELLXGENE.API = {
-        prefix: `http://localhost:<%= CXG_SERVER_PORT %>${location.pathname}/api/`,
+        prefix: `http://localhost:<%= CXG_SERVER_PORT %>${location.pathname.replace(/\/$/, "")}/api/`,
         version: "v0.3/",
       };
     </script>
--- a/client/index.html
+++ b/client/index.html
@@ -46,6 +46,9 @@
     <script type="text/javascript">
       window.CELLXGENE = {};
       window.CELLXGENE.API = {
+        // When there's a trailing slash on location.pathname, local e2e tests fail because
+        // they attempt to access http://localhost:3000//pbmc3k.cxg/api/ instead of 
+        // http://localhost:3000/pbmc3k.cxg/api/. This is a workaround for that issue.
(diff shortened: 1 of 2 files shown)


client/index.html
```diff
@@ -46,7 +46,7 @@
     <script type="text/javascript">
       window.CELLXGENE = {};
       window.CELLXGENE.API = {
-        prefix: `http://localhost:<%= CXG_SERVER_PORT %>${location.pathname}/api/`,
+        prefix: `http://localhost:<%= CXG_SERVER_PORT %>${location.pathname.replace(/\/$/, "")}/api/`,
```
tihuan, line 49: let's add a comment here to explain why we need to explicitly remove `/` and perhaps point to the Slack thread for future context!

seve: Slack threads aren't forever :/

client/src/globals.ts
```diff
@@ -27,6 +27,7 @@ export const QUERY_PARAM_EXPLAIN_NEW_TAB = "explainNewTab";
  * and ending with "api". This setup ensures it starts capturing at the single-character dataroot and includes any optional subpaths.
  */
 const REGEX_PATHNAME = /(?<=\/)[^/]{1}\/(?:[^/]+\/)*[^/]+\/(?=api)/;
+const REGEX_PATHNAME_FOR_DATASET_SELECTOR = /(?<=\/)\w+\/[^/]+\/(?=api)/;
```
tihuan, line 30: maybe add a comment to explain the context why we need a specific REGEX for dataset selector instead of reusing the REFEX above?

tihuan (approved):
LGTM after adding the comments to capture context for the code 🥇 Thanks so much, Alec 🙏 !!

client/index.html
tihuan: Oh right, not even the public channel.. good call. Maybe just copy some tl;dr and put it in the PR description as context?

client/src/globals.ts
tihuan: Great explanation! Thank you!

Issue #393, opened 2022-08-12
Enable download png of explorer image


signechambers1:
Enabled currently in excellxgene, this ticket captures the work to enable users to download an image of explorer.

seve:
dupe: #728

seve:
Duplicate of #728

PR #817, opened 2024-03-08
feat: enable image capture and download of graph

seve:
#393 

Adds image downloading button

Feature flag: `?dl=true`

Also adds some QOL's for local dev and testing


kaloster (approved):
This looks good to me! 🚀 

More of a product thing but maybe we should add some branding to the downloaded image

spatial_emb

seve:
>More of a product thing but maybe we should add some branding to the downloaded image

Oh god please no

PR #830, opened 2024-03-15
Wait for ECS deployment

ebezzi:
CI/CD change to ensure that e2e tests run after the deployment has fully completed.


.github/workflows/deploy-happy-stack.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           env: ${{ env.DEPLOYMENT_STAGE }}
           env-file: /home/runner/work/custom/envfile
           happy_version: "0.79.1"
```
atolopko-czi, line 56: worth upgrading happy now as well? (is there a chance a later version of `happy` will block until the deployment is finished?)

ebezzi: Not sure if that feature has ever been considered.

Issue #823, opened 2024-03-12
Export image downloads twice


tihuan:
Context:
1. Clicking on the camera button once downloads the same image files twice

DoD:
1. It should only download one image file

PR #835, opened 2024-03-18
fix: add screenCap to cached props for diffing during render check

seve:
Added some checks to see if screenCap differs from last attempt to re-render before actually triggering re-render.

Also adds new spatial dataset to example-datasets

These changes prevent the image download flow from download multiple images when pressed a single time while the spatial feature flag is enabled

#823


client/src/components/graph/graph.tsx
```diff
@@ -863,9 +866,15 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
-      asyncProps;
-    const { screenCap } = this.props;
+    const {
+      positions,
+      colors,
+      flags,
+      height,
+      width,
+      imageUnderlay,
+      screenCap,
```
tihuan, line 876: just wanted to double check, `screenCap` was coming from `this.props`, but now we're getting it from `asyncProps`. Is that intentional? Thanks!

example-dataset/super-cool-spatial.cxg/Xc/__commits/__1708035189525_1708035193898_3917086ff41e413989ed3d822fc83b1f_18.wrt
tihuan, line 1: these files can all be deleted?

tihuan: Ah nvm, this is the new spatial dataset. Thanks for adding this, Seve!

tihuan:
@seve can you help putting down a description of the bug, so the reviewers have more context please? Thank you!

client/src/components/graph/graph.tsx
seve: it's not necessary to get it from asyncProps here, both places are correct. The reason we're adding screenCap to watchProps is to piggyback on the prop caching.

tihuan:
@seve PTAL this approach! Would love to see what you think 🙏 

https://github.com/chanzuckerberg/single-cell-explorer/compare/thuang-fix-double-download?expand=1

Thank you!

tihuan (approved):
LGTM!! Thanks for the great fix, @seve 🚀 👏 !!

Issue #843, opened 2024-03-20
[PROD] Lasso is broken 


tihuan:
Context:

@seve Lasso isn’t resetting on failure to close the lasso

https://chanzuckerbergteam.slack.com/archives/C06FM5ZQE3Y/p1710966938965819

Repro:

1. Visit https://cellxgene.dev.single-cell.czi.technology/d/pbmc3k.cxg/?spatial=false
2. Use lasso and create an invalid lasso
3. Expect the invalid lasso to go away, but it's stuck and we cannot use lasso anymore

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFlJQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b0b4b2f5b26d2095d5dcd52f1f37ad40415de1c4/Screen%20Recording%202024-03-25%20at%208.08.20%E2%80%AFAM.mov)

DoD:
1. Fix the bug
2. Add e2e test

tihuan:
Moving this back to WIP to make sure we have e2e test. Thanks, @seve !

PR #848, opened 2024-03-25
fix: fix lasso canceling by removing definition checking

seve:
#843 

Lasso wasn't cancelling correctly because the function was always expecting the close path to be defined, even thought it almost never is.


tihuan (approved):
Thanks so much for the fix, Seve!

tihuan:
^ @seve PTAL for adding e2e test. Thank you!

PR #852, opened 2024-03-26
feat: spatial refactor

kaloster:


tihuan:
@kaloster seems like subsetting feature is actually broken, so e2e tests are correctly failing! I'm hoping once the bug is fixed all e2e tests will pass 🤞 

Thank you!

CC: @seve

client/__tests__/util/stateManager/sampleResponses.ts
```diff
@@ -75,6 +75,15 @@ const aSchemaResponse: { schema: RawSchema } = {
       obs: [{ name: "umap", type: "float32", dims: ["umap_0", "umap_1"] }],
       var: [],
     },
+    uns: {
```
tihuan, line 78: can we add a comment here to tell the readers that the response here should always match the `if spatial is None` case in `def spatial_metadata_get(spatial)` in case the response shape changes in the future?

Also add a comment in `def spatial_metadata_get(spatial)` to remind us to update `sampleResponses.ts` when the response shape changes

Thank you!

client/package.json
```diff
@@ -15,6 +15,7 @@
     "e2e-dev": "CXG_URL_BASE=https://cellxgene.dev.single-cell.czi.technology playwright test",
     "e2e-stage": "CXG_URL_BASE=https://cellxgene.staging.single-cell.czi.technology playwright test",
     "e2e-prod": "CXG_URL_BASE=https://cellxgene.cziscience.com playwright test",
+    "e2e-copy-snapshots-to-linux": "for file in __tests__/e2e/e2e.test.ts-snapshots/*-darwin.txt; do cp \"$file\" \"${file/-darwin.txt/-linux.txt}\"; done",
```
tihuan, line 18: PSA: Use `npm run e2e-copy-snapshots-to-linux` to copy updated darwin snapshots to linux equivalents!

server/tests/unit/dataset/test_cxg_dataset.py
```diff
@@ -30,6 +30,9 @@ def test_schema_filters_int64(self):
         schema = data.get_schema()
         with open(f"{FIXTURES_ROOT}/schema_without_int64.json", "r") as json_file:
             expected_schema = json.load(json_file)
+
+            # remove uns from schema comparison until test dataset is updated
```
tihuan, line 34: I might be missing something here, @kaloster 😆 Doesn't this PR update the test spatial dataset?

client/src/components/graph/graph.tsx
```diff
@@ -635,19 +631,23 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
-    const [layoutDf, colorDf, pointDilationDf] = await this.fetchData(
+    const [layoutDf, colorDf, pointDilationDf, spatial] = await this.fetchData(
       annoMatrix,
       layoutChoice,
       colorsProp,
       pointDilation
     );
     const { currentDimNames } = layoutChoice;
 
+    // This  accessor will be cleaned up post schema migration
```
tihuan, line 646: is there a ticket created to clean this up? Let's add a TODO and link the ticket here thank you!

tihuan (commented):
LGTM! Just some minor comments I can contribute lol Thanks so much for carrying the refactor through the finish line, Ronen! 👏 🙏 🚀 !!

server/tests/unit/dataset/test_cxg_dataset.py
kaloster: I think I meant we need to update the expected schema which comes from `schema_without_int64.json` which doesn't have `uns` - otherwise the test will fail

client/src/components/graph/graph.tsx
```diff
@@ -616,15 +612,15 @@ class Graph extends React.Component<GraphProps, GraphState> {
 
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
```
seve, line 613: can we drop this?

seve, line 613: Can we remove the disable?

client/src/components/graph/graph.tsx
```diff
@@ -670,12 +670,12 @@ class Graph extends React.Component<GraphProps, GraphState> {
 
     this.isSpatial = getFeatureFlag(FEATURES.SPATIAL);
 
-    // (seve): this is leading to multiple image downloads, if this isn't large, maybe we should do it as part of the initial load and store it in redux
```
seve, line 673: Can we leave this comment in?

seve (approved):
Non-blocking nits, should we create a ticket to investigate moving the dataset metadata ("corpora_props") to the uns loader rather than the config endpoint?

tihuan:
e2e all passing now 🤩 👏 !! Thanks for the amazing work, everyone!!

client/src/components/graph/graph.tsx
kaloster: it is currently stored in redux 👍🏽

client/src/reducers/controls.ts
```diff
@@ -439,7 +451,33 @@ const Controls = (
...
+      return {
+        ...state,
+        loading: true,
+        error: null,
+      };
+    case "request uns metadata success":
+      return {
+        ...state,
+        loading: false,
+        unsMetadata: action.data,
```
tihuan, line 473: Should we do the following, so any subsequent uns metadata can be preserved instead of overwritten?

I'm assuming:

```ts
datasetUnsMetadataFetchAndLoad(dispatch, "spatial"); // returns an object with one key { spatial: { ... } }

datasetUnsMetadataFetchAndLoad(dispatch, "foo"); // returns an object with one key { foo: { ... } }
```

So the following reducer will yield state:

```ts
unsMetadata: {
  spatial: { ... },
  foo: { ... }
}
```

```suggestion
        unsMetadata: { 
          ...state.unsMetadata,  // copy the current unsMetadata object
          ...action.data, // only overwrite the key/value pair from `action.data`
        },
```

server/tests/unit/common/apis/test_api_v3.py
```diff
@@ -726,6 +726,17 @@ def test_post_summaryvar_lossy(self):
...
+    def test_uns_metadata_get(self):
+        endpoint = "uns/meta"
+        query = "key=spatial"
+        for url_base in [self.TEST_URL_BASE, self.TEST_URL_BASE_SPARSE]:
+            with self.subTest(url_base=url_base):
+                url = f"{url_base}{endpoint}?{query}"
+                header = {"Accept": "application/ojson"}
+                result = self.client.get(url, headers=header)
+                self.assertEqual(result.status_code, HTTPStatus.OK)
+                self.assertEqual(result.headers["Content-Type"], "application/json")
```
tihuan, line 738: should we test the actual response content matches expected metadata?

tihuan (commented):
Looks really good! Thanks so much, Ronen 🙏 Just two small comments otherwise LGTM!

tihuan:
Don't worry about the UI Tests part! I rejected some snapshots because the current ones are better than the PR ones

client/src/components/graph/graph.tsx
```diff
@@ -614,17 +604,16 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
-    this.downloadedImg.crossOrigin = "anonymous";
-    this.downloadedImg.src = src;
+  loadTextureFromProp = async (src: string): Promise<HTMLImageElement> => {
+    this.underlayImage.crossOrigin = "anonymous";
+    this.underlayImage.src = src;
 
     await new Promise((resolve, reject) => {
-      this.downloadedImg.onload = () => resolve(this.downloadedImg);
-      this.downloadedImg.onerror = reject;
+      this.underlayImage.onload = () => resolve(this.underlayImage);
```
seve, line 612: I think I'm confused on the structure of this promise that we're creating. What exactly in here is asynchronus?

client/src/components/graph/graph.tsx
```diff
@@ -141,6 +136,7 @@ type GraphProps = Partial<RootState>;
   screenCap: state.controls.screenCap,
   mountCapture: state.controls.mountCapture,
   imageUnderlay: state.controls.imageUnderlay,
+  unsMetadata: state.controls.unsMetadata,
```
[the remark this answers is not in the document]
kaloster: The intention with this change is to create a more generic interface for `uns` data and less specific to only `spatial` data, based on past discussions with the team, and forward looking should we ever need to store and pull anything else from the `uns` array. Naming it `unsMetadata` is more descriptive, stays true to the `cxg` structure and hints on where we pulled that object from

client/src/components/graph/graph.tsx
kaloster: True - this is leftover from when we had a dedicated endpoint for this

server/tests/unit/common/apis/test_api_v3.py
kaloster: yes we should

client/src/reducers/controls.ts
kaloster: yes good call

client/src/common/types/entities.ts
```diff
@@ -62,6 +62,19 @@ export interface DatasetMetadata {
   s3_URI: S3URI;
 }
 
+/**
+ * Represents the metadata for a dataset's uns (unstructured) data.
+ * Currently only contains spatial metadata.
```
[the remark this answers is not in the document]
seve: I think my comment still stands

kaloster:
tagging @atarashansky if you could look and comment on the b/e implementation 🙏🏽

PR #859, opened 2024-03-28
fix: Correctly load image underlay when initial layout is spatial

### LINKED ISSUE #847: Image layer doesn't turn on when the image load is slow
Context:

Repro:
1. Visit https://cellxgene.dev.single-cell.czi.technology/e/spatial-4.cxg/?spatial=true
2. After the embedding is loaded, we expect the image layer to be automatically turned on, but it's not

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejhJQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1e5a674baba034415ca8dfbcb380ff3397fefa9b/Screen%20Recording%202024-03-25%20at%208.05.05%E2%80%AFAM.mov)

DoD:
1. After the embedding is loaded, we expect the image layer to be automatically turned on
2. Add E2E test for this please

seve:
The spatial underlay was only being enabled on initialization when the initial layout was flickering from one layout to spatial. This sounds like a bug. This PR introduces logic for when the spatial layout is correctly selected during the initial data load.

closes #847


tihuan (approved):
LGTM! Thanks for the fix, Seve 🚀 !!

PR #876, opened 2024-04-03
fix: remove implicit state toggling for spatial underlay in favor of explicit state setting

seve:
Old redux action handling image underlay display state was error-prone. This PR removes the reducer's implicit state toggling with explicit state sent via the action payload

### DIFF
--- a/client/src/reducers/controls.ts
+++ b/client/src/reducers/controls.ts
@@ -455,6 +455,7 @@ const Controls = (
          Uns/Spatial
     **************************/
     case "toggle image underlay": {
+      if (state.imageUnderlay === action.toggle) return state;
       return {
         ...state,
         imageUnderlay: action.toggle,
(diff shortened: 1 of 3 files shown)


kaloster (approved):
Looks good, thanks!

client/src/reducers/controls.ts
```diff
@@ -457,7 +457,7 @@ const Controls = (
     case "toggle image underlay": {
       return {
         ...state,
-        imageUnderlay: !state.imageUnderlay,
```
tihuan, line 460: let's check the existing state and only return a new state if it's different, so we don't unnecessarily trigger a re-render.

E.g.,

```ts
case "toggle image underlay": {
  if (state.imageUnderlay === action.toggle) return state;

  return { ... }
```

tihuan (commented):
Looks great! Just one optimization to do 👌 Thanks so much, Seve! 🚀

tihuan (approved):
Thanks for the update!!

PR #878, opened 2024-04-05
feat(download): include a continuous legend in image downloads

seve:
This PR allows for the inclusion of a continuous legend when downloading an image of the graph when applicable.

Also minor fix for disabling image capture state and displaying a loading spinner


client/src/components/graph/graph.tsx
```diff
@@ -1016,14 +1033,59 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+          const image64 = b64Start + svg64;
+          const img = new Image();
+          img.onload = () => {
+            ctx?.drawImage(img, 100, 100);
+          };
+          img.src = image64;
+        }
+      });
+
+      await new Promise<void>((resolve) => {
```
seve, line 1082:
```suggestion
      // need to wait for the canvas to complete draws
      await new Promise<void>((resolve) => {
```

client/src/components/graph/graph.tsx
```diff
@@ -1046,8 +1108,8 @@ class Graph extends React.Component<GraphProps, GraphState> {
           a.remove();
         }, 1000);
         dispatch({ type: "graph: screencap end" });
-        this.isDownloadingImage = false;
       }
+      this.isDownloadingImage = false;
```
seve, line 1112: this needs to be done regardless of the screencap type

client/src/components/menubar/index.tsx
```diff
@@ -369,6 +369,7 @@ class MenuBar extends React.PureComponent<{}, State> {
                 style={{
                   cursor: "pointer",
                 }}
+                loading={screenCap}
```
seve, line 372: nice little loading spinner

client/src/components/graph/graph.tsx
seve: going to commit this once reviews are done, want to batch if possible

client/src/components/graph/graph.tsx
```diff
@@ -542,7 +549,17 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
   getViewportDimensions = (): { height: number; width: number } => {
-    const { viewportRef } = this.props;
+    const { viewportRef, screenCap } = this.props;
+
+    if (screenCap) {
+      const prevAspectRatio =
+        viewportRef.clientHeight / viewportRef.clientWidth;
+
+      return {
+        height: 1800 * prevAspectRatio,
```
tihuan, line 559: I forgot why we picked 1800 instead of any other numbers 😆 

Maybe extract as constant and add comment?

client/src/components/graph/graph.tsx
```diff
@@ -1008,7 +1025,51 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+          // get style from node
+          const style = window.getComputedStyle(node as HTMLCanvasElement);
+          const nodeWidth = parseInt(style.width, 10);
+          const nodeHeight = parseInt(style.height, 10);
+          const nodeLeft = parseInt(style.left, 10);
+          const nodeTop = parseInt(style.top, 10);
+          ctx?.drawImage(
+            node as HTMLCanvasElement,
+            100 + nodeLeft,
+            100 + nodeTop,
```
tihuan, line 1048: let's add comment to explain why we need to add `100` offset

client/src/components/graph/graph.tsx
```diff
@@ -1008,7 +1025,51 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+      const offscreenCanvas = document.createElement("canvas");
+      offscreenCanvas.width = width;
+      offscreenCanvas.height = height;
+      const legendNodes =
+        document.getElementById("continuous_legend")?.childNodes;
+      const ctx = offscreenCanvas.getContext("2d");
+      ctx?.drawImage(graph, 0, 0);
+
+      legendNodes?.forEach(async (node) => {
+        if (node.nodeName === "CANVAS") {
```
tihuan, line 1038: do we need to check for `"canvas"` too like we do with `"svg" || "SVG"`?

client/src/components/graph/graph.tsx
```diff
@@ -1008,7 +1025,51 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+          );
+        } else if (node.nodeName === "svg" || node.nodeName === "SVG") {
+          const svgNode = node as SVGSVGElement;
+          const xml = new XMLSerializer().serializeToString(svgNode);
+          const svg64 = btoa(unescape(encodeURIComponent(xml)));
+          const b64Start = "data:image/svg+xml;base64,";
+          const image64 = b64Start + svg64;
+          const img = new Image();
+          img.onload = () => {
+            ctx?.drawImage(img, 100, 100);
```
tihuan, line 1060: is the 100 here linked to the offset `100` above? `100 + nodeLeft`

If so, let's use a constant to link them instead

client/src/components/graph/graph.tsx
```diff
@@ -1008,7 +1025,51 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+          };
+          img.src = image64;
+        }
+      });
+
+      await new Promise<void>((resolve) => {
+        setTimeout(() => {
+          resolve();
+        }, 25);
```
tihuan, line 1069: just curious does timeout `0` work here? If so, it's probably easier to understand that we just want to execute the download at the next event loop instead of an arbitrary 25ms wait which I imagine is both CPU/GPU dependent and thus likely lead to bugs on different computers?

tihuan (commented):
Just some quick comments 👍 Thanks for adding this, @seve !!

client/src/components/graph/graph.tsx
seve: lol i miss typed 1080, are we cool with this magic number?

client/src/components/graph/graph.tsx
seve: it actually didn't!

seve: I tried that 0 after 100 and 25 was the next number i tried

client/src/components/graph/graph.tsx
```diff
@@ -542,7 +549,18 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+    const { viewportRef, screenCap } = this.props;
+
+    if (screenCap) {
+      const prevAspectRatio =
+        viewportRef.clientHeight / viewportRef.clientWidth;
+
+      // seve: Default to 1080p resolution (arbitrary, but a good starting point for screen captures)
+      return {
+        height: 1080 * prevAspectRatio,
+        width: 1080,
```
kaloster, line 561: Possible to dynamically determine the user's screen resolution like this, using the `screen` object, instead of hardcoding it?

```
const screenWidth = screen.width;
const screenHeight = screen.height;
```

kaloster (commented):
Looks great @seve ! 

Just one minor optimization suggestion 🙏🏽

client/src/components/graph/graph.tsx
seve: we aren't programatically determining the user's screeen size. We're grabbing the aspect ratio of the viewport which the graph is rendered, and maintaining that aspect ratio while setting the width to a constant size.

kaloster (approved):

PR #881, opened 2024-04-08
fix: spatial test and dataset

kaloster:
Update to failed test: expand gene and brush with new values

The failing test was due to updated alignment caused by the `crop_box()` method introduced in the refactor PR to maintain 1:1 aspect ratio. That changed the number of cells captured in the subset and got the test to fail.

Also updated the `super-cool-spatial.cxg` version both in s3 and in repo


seve (approved):

client/__tests__/e2e/data.ts
```diff
@@ -321,8 +321,8 @@ export const datasets = {
       withSubset: "69",
     },
     expandGeneAndBrush: {
-      default: "52",
-      withSubset: "49",
```
seve, line 325: Looks like the smoke test is getting the old expected value (52)

kaloster: yes! thank you

Issue #871, opened 2024-04-02
test: Verify that image download is working properly


seve:
Related features: #728, #855, #853 

Add snapshot testing of the downloaded image and the color by legends

Issue #855, opened 2024-03-27
Export Image: Add color by legends for categorical metadata


seve:
Currently, downloads contain no information about what the `colorby` value is.  In order to allow users to include a legend in their downloads, we should download the LSB as a seperate image

Related ticket: #728

PR #889, opened 2024-04-17
test: Create chromatic snapshot testing for image downloads

seve:
#871 

Adds some testing to ensure that image downloads are working properly

### DIFF
--- a/client/__tests__/e2e/e2e.test.ts
+++ b/client/__tests__/e2e/e2e.test.ts
@@ -1042,7 +1042,7 @@ for (const testDataset of testDatasets) {
         }
       }
 
-      test.only("with continuous legend", async ({ page }, testInfo) => {
+      test("with continuous legend", async ({ page }, testInfo) => {
         await goToPage(page, url);
 
         await addGeneToSearch("SIK1", page);
--- a/client/__tests__/e2e/e2e.test.ts
+++ b/client/__tests__/e2e/e2e.test.ts
@@ -1045,13 +1045,16 @@ for (const testDataset of testDatasets) {
         const imageFile = await fs.readFile(path, { encoding: "base64" });
 
         // attach the image at path to the dom so we can snapshot it
+        // ensure that the image is rendered on top of the graph
         await page.evaluate((imgData) => {
           const img = document.createElement("img");
           img.id = "downloaded-image";
           img.src = `data:image/png;base64,${imgData}`;
-          img.style.width = "100%";
-          img.style.height = "100%";
+          img.style.height = "100vh";
           img.style.zIndex = "1000";
+          img.style.background = "white";
+          img.style.position = "absolute";
+          img.style.top = "0";
           document.body.appendChild(img);
         }, imageFile);
         await takeSnapshot(page, info);
--- a/client/__tests__/e2e/e2e.test.ts
+++ b/client/__tests__/e2e/e2e.test.ts
@@ -11,8 +11,8 @@
 import { Page, TestInfo } from "@playwright/test";
 import { test, expect, takeSnapshot } from "@chromatic-com/playwright";
 import os from "os";
-import fs from "fs/promises";
-import crypto from "crypto";
+// import fs from "fs/promises";
+// import crypto from "crypto";
 
 import { getElementCoordinates, tryUntil } from "./puppeteerUtils";
 import mockSetup from "./playwright.global.setup";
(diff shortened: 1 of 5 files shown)


seve:
@tihuan it looks like the screenshotting isn't working as I thought/intended it to. It's still capturing the original page, even though I'm navigating to a new page. I can confirm that locally the navigation is working correctly.

tihuan:
Huh interesting! What does the code look like? Should we open a new tab instead to be super sure Playwright is not on the original page URL?

Maybe ChatGPT can help us here 😁

tihuan:
I linked this PR to https://github.com/chanzuckerberg/single-cell-explorer/issues/855 is that right? Thanks!

client/__tests__/e2e/e2e.test.ts
```diff
@@ -1017,6 +1035,80 @@ for (const testDataset of testDatasets) {
...
+        await graphDownload.saveAs(graphPath);
+
+        if (legendPromise) {
+          const legendDownload = await legendPromise;
+          const legendPath = `${tmp}/${safeTitle}/${legendDownload.suggestedFilename()}`;
+          await legendDownload.saveAs(legendPath);
+
+          await navigateToAndSnapshotImage(page, info, legendPath);
+        }
+        await navigateToAndSnapshotImage(page, info, graphPath);
```
tihuan, line 1092: Do you think we can combine the `if (info.title.includes("categorical")) {` and `if (legendPromise) {` blocks like so?

```suggestion      
        await graphDownload.saveAs(graphPath);

        if (info.title.includes("categorical")) {
          const legendDownload = await page.waitForEvent("download");
          const legendPath = `${tmp}/${safeTitle}/${legendDownload.suggestedFilename()}`;
          await legendDownload.saveAs(legendPath);

          await navigateToAndSnapshotImage(page, info, legendPath);
        }
        await navigateToAndSnapshotImage(page, info, graphPath);
```

client/__tests__/e2e/playwright.global.setup.ts
```diff
@@ -43,11 +43,20 @@ const setup = async ({ page }: { page: Page }) => {
...
+
     if (message.type() === "error") {
       throw new Error(`CLIENT SIDE ERROR: ${JSON.stringify(message)}`);
     }
   });
   page.on("pageerror", (error) => {
+    // if this is running locally don't fail on error
+    if (process.env.CI !== "true") {
+      return;
+    }
```
tihuan, line 59: ahh good solution! Thanks for adding this!

tihuan (commented):
Thanks for solving this hard problem, Seve!! PTAL some small nits otherwise LGTM 🚀 🚀 🚀

client/__tests__/e2e/e2e.test.ts
```diff
@@ -89,6 +92,21 @@ const geneToRequestInfo = "SIK1";
...
+//   "CELLxGENE_legend.png e2e/e2e.test.ts,dataset: pbmc3k.cxg,Image Download,with categorical legend hash":
+//     "454f80c63655f9b4a161b0e6c5c7400c",
+//   "CELLxGENE_legend.png e2e/e2e.test.ts,dataset: super-cool-spatial.cxg,Image Download,with categorical legend hash":
+//     "179fa82ba200d4307bd41ec16d62bcef",
+//   "CELLxGENE_spatial_emb.png e2e/e2e.test.ts,dataset: super-cool-spatial.cxg,Image Download,with categorical legend hash":
+//     "9d7f861e5e953091af26046e6e82890b",
+//   "CELLxGENE_spatial_emb.png e2e/e2e.test.ts,dataset: super-cool-spatial.cxg,Image Download,with continuous legend hash":
+//     "59d6d8b5d5eadc79387fa74bd28683eb",
+// };
+
```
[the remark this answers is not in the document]
seve: forgot to delete!
```suggestion
```

client/__tests__/e2e/e2e.test.ts
```diff
@@ -545,8 +563,8 @@ for (const testDataset of testDatasets) {
 
           expect(initialCount).toBe(panzoomLasso.count);
 
-          await page.getByTestId("mode-pan-zoom");
-          await page.getByTestId("mode-lasso");
+          page.getByTestId("mode-pan-zoom");
+          page.getByTestId("mode-lasso");
```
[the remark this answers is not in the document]
seve: Not sure either, this was just a result of me removing unnecessary awaits

client/__tests__/e2e/e2e.test.ts
seve: I'm going to leave it as is for now. Playwright's Download API isn't well documented, especially in multiple download cases. To be clear, this is the functionality we're watching:

1.  Download listener is initiated
2.  Click button that results in image capture and download
3.  First download is initiated and captured by listener
4.  We prepare a new download listener
5.  Playwright saves the download using `.saveAs`, my understanding is  this is similar, if not exactly, to the behavior of choosing where to download the file in the modal
6.  Once the first download is confirmed and saved, the second download (legend) is fired
7.  Playwright captures download step 5 repeats

I'm afraid that if we created and awaited the download listener after step 5, it would timeout as we missed the download event

client/__tests__/e2e/e2e.test.ts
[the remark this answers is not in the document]
kaloster: I think these are either preparation for something else or if the intention is to just verify the visibility then it should be updated to:

```
await expect(page.getByTestId("mode-pan-zoom")).toBeVisible();
await expect(page.getByTestId("mode-lasso")).toBeVisible();
```

Otherwise these are no-op and can be safely removed in my opinion

client/__tests__/e2e/e2e.test.ts
kaloster: my 2 cents on this - I believe Timmy's suggested edit cleans it up and simplifies it - not sure we need the extra `if (legendPromise)` but also understand your cautiousness around not missing the `download` event. 

Can we try to just trigger both downloads based on the first `const graphPromise = page.waitForEvent("download");` as we are repeating this await for the same event? 

Then move this to after the condition block of the legend/categorical:

```suggestion
        if (info.title.includes("categorical")) {
          // this will capture a separate legend download if the colorby is categorical
          legendPromise = page.waitForEvent("download"); // is this needed?
        }
     
        if (legendPromise) {
          const legendDownload = await legendPromise;
          const legendPath = `${tmp}/${safeTitle}/${legendDownload.suggestedFilename()}`;
          await legendDownload.saveAs(legendPath);

          await navigateToAndSnapshotImage(page, info, legendPath);
        }
        await graphDownload.saveAs(graphPath); // move graph download here
        await navigateToAndSnapshotImage(page, info, graphPath);
```

seve: This won't work because the graph download needs to be saved in order for the legend download to be triggered. In your suggestion, the legend await will hang or capture the initial graph download. Step 6 described above is crucial

client/__tests__/e2e/e2e.test.ts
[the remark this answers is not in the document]
seve: took a look at the test and looks like its been broken for some time. Fixed it

client/__tests__/e2e/e2e.test.ts
seve: (i'm actually testing this solutions that are being recommended and they don't work)

kaloster: Ok so that's why it was a suggestion because you're the one who's actually running the tests 😁  Then I'd say keep it in the state that works just maybe add a comment explicitly mentioning the need for step 6 in that sequence of creating the second event listener before the first saveAs

kaloster (approved):
LGTM! 🚀

client/__tests__/e2e/e2e.test.ts
tihuan: Ahh got it! Thanks for the explanation, Seve! SGTM 👍

PR #902, opened 2024-04-26
chore: add e2e tests for cellguide cxgs and related features

atarashansky:
Adding E2E tests for the following interactions:
 - Assert no author categories or standard categories - done
 - Marker Gene Sets is visible and collapsed. Expansion reveals expected gene sets - done
 - Gene sets are alphabetically sorted. - done
 - Gene set expansion works as expected. - done
 - Gene set modification/deletion and page refresh resets the gene set - done
 - Categorical labels are sorted in descending order by mean expression when genesets are colored by. - done
 - Labels with 0 cells after subsetting are filtered out. - done

Added a test fixture in the `cellguide-cxgs/` path to set `isCellGuideCxg` equal to true.

Note that because we now have test coverage for the new categorical label behaviors, I removed the isCellGuideCxg feature flag from it. Now all explorer instances will benefit from the improved UX. Namely, the improvements are:
 - Remove labels with 0 cell counts when subsetting
 - Sort labels in descending order when coloring by gene expression or any other continuous metadata.

Also, to reduce the size of the new test fixture added (`example.cxg`), I set the gene expression tiledb arrays `Xr` and `Xc` to have no data in it. This is fine as none of the cellguide-specific tests require gene expression data.


client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -405,11 +429,10 @@ export async function checkGenesetDescription(
 
       const editButton = `${genesetName}:edit-genesetName-mode`;
       await page.getByTestId(editButton).click({
-        force: true
+        force: true,
         /**
          * (thuang): Don't wait for the default timeout, since we want to fail fast
-         */,
-        timeout: 1 * 1000,
+         */ timeout: 1 * 1000,
```
tihuan, line 435: nit: just curious, is it auto-formatting that put the comment and the code on the same line?
```suggestion
         */ 
         timeout: 1 * 1000,
```

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -140,6 +140,30 @@ export async function getAllCategoriesAndCounts(
...
+    categoryRows.map(async (row): Promise<string> => {
+      const cat = await row
+        .getByTestId("categorical-value")
+        .getAttribute("aria-label");
+
+      if (!cat) throw new Error("category value not found");
+
+      return cat;
+    })
+  );
```
tihuan, line 163: discussion: Not sure how flaky this is. Is it worth adding a `tryUntil()` to ensure `arrayOfLabels` all resolve properly before returning it?

E.g.,

```ts
const categoryRows = await page
    .getByTestId(`category-${category}`)
    .getByTestId("categorical-row")
    .all();

// this retries until we get all the labels 
await tryUntil(async () => {
  await Promise.all(
    categoryRows.map(async (row): Promise<string> => {
      const cat = await row
        .getByTestId("categorical-value")
        .getAttribute("aria-label");

      if (!cat) throw new Error("category value not found");

      return cat;
    })
  );
}, { page })

const arrayOfLabels = await Promise.all(
    categoryRows.map(async (row): Promise<string> => {
      const cat = await row
        .getByTestId("categorical-value")
        .getAttribute("aria-label");

      if (!cat) throw new Error("category value not found");

      return cat;
    })
  );

return arrayOfLabels;
```

client/__tests__/e2e/e2e.test.ts
```diff
@@ -101,6 +104,143 @@ const testURLs = {
...
+        );
+        const geneset2 = await page.locator(
+          `div[data-testid="geneset"]:has-text("smooth muscle fiber of ileum")`
+        );
+
+        await expect(geneset1).toBeVisible();
+        await expect(geneset2).toBeVisible();
+      },
+      { page }
+    );
```
tihuan, line 176: discussion: I'm thinking the click should be part of the tryUntil, in case the click is not registered and the section is never expanded, then the tryUntil will never succeed

client/__tests__/e2e/e2e.test.ts
```diff
@@ -101,6 +104,143 @@ const testURLs = {
   "super-cool-spatial.cxg": pageURLSpatial,
 };
 
+describe(`Testing CellGuideCXG at ${pageURLCellGuide}`, () => {
```
tihuan, line 107: Maybe we should put the CellGuideCXG test in its own test file?

client/__tests__/e2e/e2e.test.ts
```diff
@@ -101,6 +104,143 @@ const testURLs = {
...
+      "svg[data-icon='chevron-down']"
+    );
+    if ((await chevronDownIcon.count()) === 0) {
+      await markerGeneSetsHeader.click();
+    }
+
+    // Locate and expand 'enteric smooth muscle cell' geneset
+    await expandGeneset("enteric smooth muscle cell - marker genes", page);
+
+    // Assert the presence of the genes div and that it contains 100 child divs
```
tihuan, line 197: Why 55 child divs? Why does it have to be 55? 

Just thinking about test durability to reduce flakiness in case of data change, do we just need to assert that there's > 0 child?

```suggestion
    // Assert the presence of the genes div and that it contains 55 child divs
```

client/__tests__/e2e/e2e.test.ts
```diff
@@ -101,6 +104,143 @@ const testURLs = {
...
+    // Locate and expand the 'Marker Gene Sets' header if not already expanded
+    markerGeneSetsHeader = await page.locator(
+      "h5:has-text('Marker Gene Sets')"
+    );
+    chevronDownIcon = markerGeneSetsHeader.locator(
+      "svg[data-icon='chevron-down']"
+    );
+    if ((await chevronDownIcon.count()) === 0) {
+      await markerGeneSetsHeader.click();
+    }
```
tihuan, line 235: same code as above, let's extract this as a function to keep the code DRY 🙏 

Also might be worth using `tryUntil()` to check that after `await markerGeneSetsHeader.click();` the app is in the expected state (i.e., `(await chevronDownIcon.count()) === 1` ), in case that the click is not registered

tihuan (commented):
Looks really good! Thanks for adding comprehensive tests, Alec 🙏 Some v minor comments otherwise LGTM 🔥 !!

client/__tests__/e2e/e2e.test.ts
atarashansky: I moved the click outside because I thought it would be bad to toggle it open/closed each iteration of tryUntil, but since it runs a bunch of times I guess i would be fine.

client/__tests__/e2e/e2e.test.ts
atarashansky: There's no data change since it's a local fixture. There's 55 because there are 55 gene sets in the fixture :')

client/__tests__/e2e/cellxgeneActions.ts
atarashansky: I don't think it's flakey.

client/__tests__/e2e/e2e.test.ts
atarashansky: Not sure how to config playwright to run the separate file - if you think it's important, perhaps you could commit it to this branch so I can learn? 👼

client/__tests__/e2e/e2e.test.ts
atarashansky: Wrapping it in a tryUntil caused the test to bug out, not sure why, couldn't figure it out.

client/__tests__/e2e/cellxgeneActions.ts
```diff
@@ -141,6 +141,43 @@ export async function getAllCategoriesAndCounts(
...
+export async function expandMarkerGeneSetsHeader(page: Page): Promise<void> {
+  // Locate and expand the 'Marker Gene Sets' header if not already expanded
+  const markerGeneSetsHeader = await page.locator(
+    "h5:has-text('Marker Gene Sets')"
+  );
+  const chevronDownIcon = markerGeneSetsHeader.locator(
+    "svg[data-icon='chevron-down']"
+  );
+  if ((await chevronDownIcon.count()) === 0) {
+    await markerGeneSetsHeader.click();
```
tihuan, line 153: I see that the assertion count > 0 got removed between the latest commit and the one before. Why is that? Thank you!

atarashansky: https://github.com/chanzuckerberg/single-cell-explorer/pull/902#discussion_r1583500316

atarashansky: Oh, I just realized why the tryUntil wasn't working, lol. Had to await it.

atarashansky: Whoopsie

atarashansky: Reverted the commit

tihuan (approved):
LGTMMM! Thanks so much again, Alec 👏 🙏 🔥

PR #935, opened 2024-05-17
fix: Export image update

tihuan:
1. Extract image download logics into its own method `handleImageDownload()`
2. Update `handleImageDownload()` to turn both openseadragon and graph canvases into png, and then draw on offscreen canvas, then convert to another PNG, then download

Demo:
https://github.com/chanzuckerberg/single-cell-explorer/assets/6309723/67acff82-4d9f-4db9-a29e-347db8bf345a

Download legend too:

https://github.com/chanzuckerberg/single-cell-explorer/assets/6309723/83b5d16d-d00f-461a-b583-61b712991bc6

### DIFF
--- a/client/src/components/graph/graph.tsx
+++ b/client/src/components/graph/graph.tsx
@@ -1,6 +1,5 @@
 import React, { MouseEvent, MouseEventHandler } from "react";
 import * as d3 from "d3";
-import { toPng } from "html-to-image";
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- used as @connect
 import { connect, shallowEqual } from "react-redux";
 import { mat3, vec2 } from "gl-matrix";
@@ -40,14 +39,15 @@ import { LassoFunctionWithAttributes } from "./setupLasso";
 import { Field } from "../../common/types/schema";
 import { Query } from "../../annoMatrix/query";
 import { DatasetUnsMetadata } from "../../common/types/entities";
-import { LayoutChoiceState } from "../../reducers/layoutChoice";
 
 import {
   captureLegend,
   createModelTF,
   createProjectionTF,
+  downloadImage,
   getSpatialPrefixUrl,
   getSpatialTileSources,
+  loadImage,
   shouldShowOpenseadragon,
 } from "./util";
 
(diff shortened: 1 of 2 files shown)


client/src/components/graph/graph.tsx
```diff
@@ -1,6 +1,5 @@
 import React, { MouseEvent, MouseEventHandler } from "react";
 import * as d3 from "d3";
-import { toPng } from "html-to-image";
```
tihuan, line 3: no need for this, since canvas element's toDataURL can generate PNG

client/src/components/graph/graph.tsx
```diff
@@ -69,25 +69,6 @@ interface GraphAsyncProps {
   screenCap: boolean;
 }
 
-function downloadImage(
```
tihuan, line 72: extract this to `util` file

client/src/components/graph/graph.tsx
```diff
@@ -557,6 +538,122 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+      if (mountCapture) {
+        this.setState({ testImageSrc: graphImageURI });
+        dispatch({ type: "test: screencap end" });
+      } else {
+        downloadImage(graphImageURI, layoutChoice);
+
+        let categoricalLegendImageURI: string | null = null;
+
+        // without this, the legend is drawn offscreen
+        const PADDING_PX = 100;
```
kaloster, line 631: Does the canvas width and height change according to the user's screen resolution and window resize state? If so does it make sense for this to be a percentage value instead of arbitrary value?

kaloster (approved):
LGTM! I had just one comment/question

client/src/components/graph/graph.tsx
tihuan: ooh this is a great question! Since @seve added it originally let's wait for him to answer on this one. But my gut feeling is that since this constant only affects the legend, which occupies the same width regardless of viewport size, so I'm thinking it should be okay

I'll merge the PR first and then submit a subsequent PR to fix if Seve thinks we need to update anything

Thanks so much for the quick review!

Issue #940, opened 2024-05-21
BUG: Hide image toggle button when image fetch fails


tihuan:
Context:
1. In a meeting, we decided that we should hide the image toggle button when deep zoom image fetch fails

DoD:
1. Move `isDeepZoomSourceValid` state to reducer, so MenuBar can hide image toggle button as needed

tihuan:
merged into feature branch

PR #941, opened 2024-05-21
fix: 940 hide image toggle

tihuan:
fix: #940

1. Move state `isDeepZoomSourceValid` into `config` reducer, so both Graph and Menubar components can use that state
2. Extract selector `shouldShowOpenseadragon` to `client/src/common/selectors.ts` to be used by the components above
3. Create `config` selector file

https://github.com/chanzuckerberg/single-cell-explorer/assets/6309723/7d12238b-e172-46f4-9d94-4decc71f3490

### DIFF
--- a/client/src/actions/config.ts
+++ b/client/src/actions/config.ts
@@ -1,0 +1,5 @@
+export function fetchDeepZoomImageFailed() {
+  return {
+    type: "fetchDeepZoomImageFailed",
+  };
+}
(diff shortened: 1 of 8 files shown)


tihuan:
@kaloster I think Seve doesn't have time to review this PR, so PTAL thank you 🙏 !!

kaloster (approved):

PR #944, opened 2024-05-23
pr review

seve:


client/src/reducers/layoutChoice.ts
```diff
@@ -29,16 +28,6 @@ const LayoutChoice = (
   nextSharedState: RootState
 ): LayoutChoiceState => {
   switch (action.type) {
-    case "initial data load complete": {
```
seve, line 32: no longer need to hook onto this action!

client/src/reducers/layoutChoice.ts
```diff
@@ -65,10 +54,3 @@ function getCurrentLayout(
 
   return { current: layoutChoice, currentDimNames };
 }
-
-export function getBestDefaultLayout(layouts: Array<string>): string {
```
seve, line 69: moved to callsite

client/src/actions/embedding.ts
```diff
@@ -61,3 +62,39 @@ export const layoutChoiceAction: ActionCreator<
       annoMatrix,
     });
   };
+
+export const initialLayoutChoiceAction: ActionCreator<
```
seve, line 66: New action for initial load

client/src/actions/index.ts
```diff
@@ -234,27 +232,12 @@ const doInitialDataLoad = (): ((
...
-      const availableLayouts = selectAvailableLayouts({ annoMatrix });
-      const fallbackLayout = getBestDefaultLayout(availableLayouts);
-      const defaultLayout = config?.parameters?.default_embedding || "";
-
-      const finalLayout = availableLayouts.includes(defaultLayout)
-        ? defaultLayout
-        : fallbackLayout;
-
+      // could possibly integrate this into the generation of the initial annoMatrix as part of the annoMatrix generation instead of after initialization is complete
+      embActions.initialLayoutChoiceAction(config);
```
seve, line 236: Single explicit action call in `doInitialDataLoad()`

client/src/actions/embedding.ts
```diff
@@ -61,3 +62,39 @@ export const layoutChoiceAction: ActionCreator<
...
+      prevCrossfilter,
+      finalLayout
+    );
+
+    dispatch({
+      type: "set layout choice",
+      layoutChoice: finalLayout,
+      obsCrossfilter,
+      annoMatrix,
+    });
```
seve, line 92: This could just be a call of the above action, but I think it would be more clear if each action has its own explicit dispatch

client/src/reducers/layoutChoice.ts
```diff
@@ -29,16 +28,6 @@ const LayoutChoice = (
   nextSharedState: RootState
 ): LayoutChoiceState => {
   switch (action.type) {
-    case "initial data load complete": {
-      const { layoutChoice } = action;
-
-      return {
-        ...state,
-        available: selectAvailableLayouts(nextSharedState),
```
tihuan, line 37: I think we might still need to update `available` in the state, since nowhere else is adding it 😄 ?

client/src/actions/embedding.ts
tihuan: agreed!

client/src/util/stateManager/viewStackHelpers.ts
```diff
@@ -109,6 +109,14 @@ export function _userResetSubsetAnnoMatrix(
   return annoMatrix;
 }
 
+const isAnnoMatrixClipView = (
```
tihuan, line 112: did you mean to add this file change to this PR? I might be missing something lol

client/src/reducers/layoutChoice.ts
seve: right! fixing that now

client/src/util/stateManager/viewStackHelpers.ts
seve: I was just peering into this file while reading your debugging findings and thought typing it would be helpful, lol. Not sure when else i would commit this

PR #952, opened 2024-05-28
fix: updated super cool spatial

kaloster:


tihuan (approved):
Thank you for updating it!!

PR #953, opened 2024-05-28
fix: exclude specific ontology term id keys

kaloster:
#950 

added const: ONTOLOGY_TERM_ID_KEYS
removed: ONTOLOGY_KEY
revised: isCategoryNameOntologyKey, buildDatasetMetadataViews

also - moved `in_tissue` to Standard Categories`

Screenshot 2024-06-06 at 5 39 50 PM


client/src/globals.ts
```diff
@@ -15,8 +15,17 @@ export const unassignedCategoryLabel = "unassigned";
 /** Maximum number of cells a dataset can have in order to be included for display. */
 export const DATASET_MAX_CELL_COUNT = 2_000_000;
 
-/* Category name suffix used to determine if category name is ontology term id. */
-export const ONTOLOGY_KEY = "ontology_term_id";
+/* Explicit ontology term ids. */
```
seve, line 18: why do we want to make this explicit?

kaloster: yes - the feedback from Jason was that the general filter for `ontology_term_id` was too broad and we should exclude the ones on this list based on the schema definition

seve (approved):

Issue #947, opened 2024-05-23
feat(side by side embedding): UI/UX Stretch Goals


seve:
### Stretch
- [ ] Move old panels into new side sub-panel
  - [ ] move gene info
  - [ ] move dataset info
    ![image](https://github.com/chanzuckerberg/single-cell-explorer/assets/8716829/f242406e-4ac3-41e8-8482-4fb9b96f0278)

Figma:
https://www.figma.com/design/eJ3acPxG04PKVlom6eS8ft/Spatial?node-id=0-1&t=yAWU6j6wSLKV9eIJ-0

seve:
@kaloster This ticket can be broken out from the side by side work and is unblocked if you have room to add to next sprint

PR #963, opened 2024-06-04
feat: new info panel

kaloster:
#947 

UI updated to accommodate side by side for spatial:
- Removed dataset drawer
- Moved dataset info to new info panel
- Moved gene info to new info panel
- New panel can be minimized/maximized and closed
- Revised and fixed tests

Cleaned up unused code and redux actions

https://github.com/chanzuckerberg/single-cell-explorer/assets/7976579/11b62b25-bfbb-457f-99f2-18e73fae0105


client/src/components/geneExpression/infoPanel/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+import DatasetInfo from "./datasetInfo";
+
+interface Props {
+  activeTab: string;
+  dispatch: AppDispatch;
+  infoPanelMinimized: boolean;
+  infoPanelHidden: boolean;
+}
+
+const InfoPanel = (props: Props) => {
```
tihuan, line 23: super nit: Can we change this from function expression to function declaration please 🙏 ? I believe stack tracing works a lot better with declaration than an anonymous function!

Here and throughout

E.g.,

```ts
function InfoPanel(props: Props) {
  ...
}
```

client/src/components/geneExpression/infoPanel/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  InfoPanelContent,
+  InfoPanelHeader,
+  InfoPanelTabs,
+  InfoPanelWrapper,
+  StyledAnchorButton,
+} from "./style";
+import GeneInfo from "./geneInfo";
+import DatasetInfo from "./datasetInfo";
+
+interface Props {
```
tihuan, line 16: Non-blocking: Maybe as a post launch P0 - worth refactoring types out to its own file like we do in Data Portal 🙆‍♂️ ?

client/src/components/geneExpression/infoPanel/geneInfo/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  display: "-webkit-box",
+                  WebkitLineClamp: "7",
+                  WebkitBoxOrient: "vertical",
+                  overflow: "hidden",
+                }}
+                data-testid="gene-info-summary"
+              >
+                {geneSummary}
+              </Content>
+            )}
```
tihuan, line 135: wonder if there's opportunities to keeping the code DRY by reusing the same styling object?

client/src/components/geneExpression/infoPanel/geneInfo/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  sdsType="static"
+                />
+                <span>
+                  NCBI didn&apos;t return an exact match for this gene.
+                </span>
+              </WarningBanner>
+            ) : null}
+            <GeneSymbol data-testid="gene-info-symbol">{gene}</GeneSymbol>
+            <Content data-testid="gene-info-name">{geneName}</Content>
+            {geneSummary === "" ? (
```
tihuan, line 112: non blocking, but something worth refactoring post launch - wonder if we should extract this component out? This render function has so many conditional and nested rendering that the cognitive complexity seems high to me 😁

client/src/reducers/controls.ts
```diff
@@ -1,63 +1,14 @@
 import { AnyAction } from "redux";
 
-type Level = "top" | "bottom" | "";
-
-interface StackLevels {
-  geneLevel: Level;
-  scatterplotLevel: Level;
-}
-/* logic for minimizing and maximizing pop-ups */
-const minimizeMaximizePopUps = (
```
tihuan, line 10: really awesome cleanup! Thank you!

client/src/components/geneExpression/gene.tsx
```diff
@@ -91,41 +91,27 @@ class Gene extends React.Component<Props, State> {
...
     const info = await actions.fetchGeneInfo(geneId, gene);
     if (!info) {
-      dispatch({
-        type: "open gene info",
-        gene,
-        url: "",
-        name: "",
-        synonyms: [],
-        summary: "",
-        infoError: "fetch gene info failed",
```
tihuan, line 120: maybe I'm missing something, but it seems like by removing this dispatch, we no longer tell user a gene info fetch as failed?

client/src/components/geneExpression/infoPanel/datasetInfo/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  allCategoryNames.forEach((catName) => {
+    const isUserAnno = schema?.annotations?.obsByName[catName]?.writable;
+    const colSchema = schema.annotations.obsByName[catName];
+    if (!isUserAnno && colSchema.categories?.length === 1) {
+      allSingleValues.set(catName, colSchema.categories[0]);
+    }
+  });
+  singleContinuousValues.forEach((value, catName) => {
+    allSingleValues.set(catName, value);
+  });
```
tihuan, line 38: non blocking - just wondering if there's an opportunity to cache the calculation, so we don't generate new objects every time, which could cause unnecessary re-rendering. But up to you!

client/src/components/geneExpression/infoPanel/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+        isMinimized={infoPanelMinimized}
+      >
+        {activeTab === "Gene" && <GeneInfo />}
+        {activeTab === "Dataset" && <DatasetInfo />}
+      </InfoPanelContent>
+    </InfoPanelWrapper>
+  );
+};
+
+export default connect((state: RootState) => ({
```
tihuan, line 99: in other places we define and use `mapStateToProps`, maybe worth doing the same here for consistency?

tihuan (commented):
Thanks for polishing and cleaning up the components in addition to adding the new side panel feature 🔥 !! Just some small comments PTAL thank you 👼 🙏 ✨

client/src/components/geneExpression/infoPanel/geneInfo/index.tsx
kaloster: I'll address this in a following PR

client/src/components/geneExpression/gene.tsx
kaloster: this was kind of a workaround to trigger empty state in the UI - no longer needed with the new revised logic introduced in `geneInfo/index.tsx`

"open gene info" will still return `infoError` if it occurs and will affect the UI representation accordingly.

client/src/components/geneExpression/index.tsx
```diff
@@ -175,7 +176,14 @@ class GeneExpression extends React.Component<{}, State> {
     const { isCellGuideCxg } = this.props;
     const { geneSetsExpanded, markerGeneSetsExpanded } = this.state;
     return (
-      <div>
+      <div
+        style={{
+          display: "flex",
+          flexDirection: "column",
+          justifyContent: "space-between",
+          padding: globals.leftSidebarSectionPadding,
```
seve, line 184: we're using a lsb global on the rsb, should we be using a different one or normailize this global?

client/src/components/leftSidebar/types.ts
```diff
@@ -0,0 +1,11 @@
+import { RootState } from "../../reducers";
+
+export interface Props {
+  scatterplotXXaccessor: string;
+  scatterplotYYaccessor: string;
```
seve, line 5: Let's grab these straight from the store! YMMV on my suggestion
```suggestion
  scatterplotXXaccessor: RootState["controls"]["scatterployXXaccessor"];
  scatterplotYYaccessor:  RootState["controls"]["scatterployYYaccessor"];
```

seve (changes requested):
There's also an open conversation in Slack regarding rebasing this branch on the featture branch for side by side

client/src/components/geneExpression/gene.tsx
```diff
@@ -91,41 +91,27 @@ class Gene extends React.Component<Props, State> {
...
-      });
-      return;
-    }
-
     dispatch({
       type: "load gene info",
       gene,
     });
 
+    dispatch({ type: "toggle active info panel", activeTab: "Gene" });
```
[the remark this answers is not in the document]
kaloster: Done!

kaloster:
rebased and merged [here](https://github.com/chanzuckerberg/single-cell-explorer/pull/972)

PR #978, opened 2024-06-13
fix(right-side-bar): lock GE height

kaloster:
Lock height of GE and Info panel so both overflow-y scroll


client/src/components/geneExpression/index.tsx
```diff
@@ -180,8 +180,10 @@ class GeneExpression extends React.Component<{}, State> {
         style={{
           display: "flex",
           flexDirection: "column",
-          justifyContent: "space-between",
+          justifyContent: "start",
           padding: globals.rightSidebarSectionPadding,
+          height: "50%",
```
seve, line 185:
```suggestion
          min-height: "50%",
```

Otherwise, if the drawer isn't open, the continuous part is still only able to grow to 50% of the height

seve (changes requested):
styling request

seve (changes requested):
Instead of setting the visibility of the info panel to hidden, we should set display to none. That way the elements don't take up any height when the info panel is not displayed

PR #989, opened 2024-06-18
test: side by side test fixes

tihuan:


client/__tests__/e2e/cell-guide.test.ts
```diff
@@ -32,10 +32,6 @@ describe("CellGuideCXG", () => {
   test("page launched", async ({ page }, testInfo) => {
     await goToPage(page, pageURLCellGuide);
 
-    const element = await page.getByTestId("header").innerHTML();
-
-    expect(element).toMatchSnapshot();
```
tihuan, line 37: Use Chromatic screenshot for better DX. Here and throughout

client/__tests__/e2e/e2e.test.ts
```diff
@@ -165,12 +157,6 @@ for (const testDataset of testDatasets) {
         for (const label of Object.keys(
           data.categorical
         ) as (keyof typeof data.categorical)[]) {
-          const element = await page
-            .getByTestId(`category-${label}`)
-            .innerHTML();
-
-          expect(element).toMatchSnapshot();
```
tihuan, line 172: removed this one and the following, since the bottom chromatic snapshot should capture them anyway

client/src/common/selectors.ts
```diff
@@ -7,9 +7,19 @@ import { FEATURES } from "../util/featureFlags/features";
...
 
-  return !!(
+  const { open, layoutChoice: panelEmbeddingLayoutChoice } =
+    panelEmbedding || {};
+
+  const isPanelEmbeddingInSpatialMode =
+    /**
+     * (thuang): If the side panel is not open, don't take its layout choice into account
+     */
+    open &&
```
tihuan, line 17: this is to resolve the bug when side panel is off, the image layer toggle button would still show up when the main panel is switched to a non-spatial embedding

client/src/components/embedding/index.tsx
```diff
@@ -81,16 +78,6 @@ const Embedding = (props: Props) => {
     await dispatch(
       actions.layoutChoiceAction(e.currentTarget.value, isSidePanel)
     );
-    if (
-      imageUnderlay &&
-      !isSidePanel &&
-      !e.currentTarget.value.includes(globals.spatialEmbeddingKeyword)
-    ) {
-      dispatch({
```
tihuan, line 89: this is not needed anymore, since the toggle state should only be controlled by the image toggle button only

client/src/components/graph/graph.tsx
```diff
@@ -329,6 +330,19 @@ class Graph extends React.Component<GraphProps, GraphState> {
...
+        if the selection tool or state has changed, ensure that the selection
+        tool correctly reflects the underlying selection.
+        */
+    if (
+      currentSelection !== prevProps.currentSelection ||
+      graphInteractionMode !== prevProps.graphInteractionMode ||
+      stateChanges.toolSVG
+    ) {
+      const { tool } = this.state;
+      this.selectionToolUpdate(stateChanges.tool ? stateChanges.tool : tool!);
```
tihuan, line 343: this needs to be added back to resolve the lasso element missing bug

PR #1011, opened 2024-06-28
fix: GE scroll and overflow

kaloster:
- Lock height of GE and info panel so both overflow-y scroll
- Show GE content when info panel is minimized


tihuan:
I see that e2e failed because of CORS, so I added back `AllowAllOrigins` in response header 👌 


Screenshot 2024-07-02 at 9 29 04 AM

PR #1015, opened 2024-07-02
feat: Remove spatial feature flag

tihuan:
⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️
DO NOT MERGE THIS PR UNTIL WE GET THE GREEN LIGHT 🚦
⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️⚠️

### DIFF
--- a/client/src/components/menubar/index.tsx
+++ b/client/src/components/menubar/index.tsx
@@ -249,7 +249,6 @@ const MenuBar = ({
   const isColoredByCategorical = !!categoricalSelection?.[colorAccessor || ""];
 
   const isTest = getFeatureFlag(FEATURES.TEST);
-  const isDownload = getFeatureFlag(FEATURES.DOWNLOAD);
 
   // constants used to create selection tool button
   const [selectionTooltip, selectionButtonIcon] = ["select", "polygon-filter"];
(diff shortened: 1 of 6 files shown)


client/src/components/menubar/index.tsx
```diff
@@ -278,25 +277,25 @@ const MenuBar = ({
                 data-testid="drawer"
               />
             </ButtonGroup>
-            {isDownload && (
```
tihuan, line 281: no more conditional rendering on the download button

client/src/components/menubar/style.ts
```diff
@@ -5,14 +5,8 @@ import { FEATURES } from "../../util/featureFlags/features";
...
 const isTest = getFeatureFlag(FEATURES.TEST);
-const isDownload = getFeatureFlag(FEATURES.DOWNLOAD);
 
-let FIRST_VERTICAL_THRESHOLD_WIDTH_PX = 655;
-if (isTest && isDownload) {
-  FIRST_VERTICAL_THRESHOLD_WIDTH_PX = 705;
-} else if (isTest || isDownload) {
-  FIRST_VERTICAL_THRESHOLD_WIDTH_PX = 685;
-}
+const FIRST_VERTICAL_THRESHOLD_WIDTH_PX = isTest ? 705 : 685;
```
tihuan, line 9: @rainandbare PTAL if my math is correct here lol I'm thinking that now we have removed `isDownload` feature flag,  it should mean that the default width has increased from 655 to 685. And then if `isTest` is `true`, we bump it up to 705?

Thank you!

rainandbare: If there is only ever one potential variable button then we can do: 

```suggestion
const FIRST_VERTICAL_THRESHOLD_WIDTH_PX = isTest ? 685 : 655;
```

rainandbare (approved):
Looks good to me! Left a suggestion with the appropriate break points.

client/src/components/menubar/style.ts
tihuan: Chatting with Suzette about this in Slack at the moment! Will share the result here later 👍

tihuan: Confirmed that the current implementation is correct! Thanks again for your help, Suzette 🎉 !

PR #1016, opened 2024-07-02
chore: test required checks

hspitzley-czi:


.github/workflows/argus-stack-staging-upsert.yaml
```diff
@@ -13,12 +10,22 @@ concurrency:
...
     runs-on: [ARM64, self-hosted, Linux]
     steps:
       - run: echo "The head branch of this PR is the release please branch"
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            staging_config:
+              - '.infra/common.yaml'
+              - '.infra/staging/values.yaml'
```
hspitzley-czi, line 24: Make sure this always runs so we can make it a required check without blocking feature branch PR merges

PR #1021, opened 2024-07-05
feat: accurate and dynamic dot size for spatial

kaloster:
- Update Explorer to use spot_diameter_fullres to calculate accurate spot size when displaying spatial embeddings
- Factor in camera distance or zoom level for the spot diameter
- Leave old calculation for non-spatial embeddings

Implementation:

- Add endpoint in the backend for uns["spatial"] metadata to be used by front end for above calculations
- Add state object
- Update regl file with calculation
- Add unit tests for added functionality


tihuan (approved):
Looks super good! Thanks so much for all the comments too 🙏 LGTM!!

PR #1024, opened 2024-07-08
chore: Add test for bug #1022 color by category

tihuan:
1. Add test to prevent regression for bug #1022


client/src/components/categorical/category/index.tsx
```diff
@@ -485,7 +490,7 @@ const CategoryHeader = React.memo(
               hide: { enabled: false },
             }}
           >
-            <AnchorButton
+            <Button
```
tihuan, line 493: clean up - anchor button is an `<a>` that look like a button, but not a `<button>` tag. In this case, we want `<button`>

kaloster (approved):

PR #1035, opened 2024-07-12
fix: Remove the wrong lasso in the other panel

tihuan:
1. Add `graphId` to lasso selection reducer state
2. Use the new `graphId` to determine whether to display the lasso element in the panel or not
3. We only display the lasso element in the panel that initiated the lasso

Screenshot 2024-07-12 at 12 01 52 PM

https://github.com/user-attachments/assets/cf2aa442-f862-4151-bccf-10c11bd8de5d

### DIFF
--- a/.infra/rdev/values.yaml
+++ b/.infra/rdev/values.yaml
@@ -2,7 +2,7 @@ stack:
   services:
     explorer:
       image:
-        tag: sha-26c7759a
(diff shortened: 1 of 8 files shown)


kaloster (approved):

tihuan:
e2e tests passed locally. Will merge after it passes in GHA!

Issue #1046, opened 2024-07-19
Opacity should only apply to spacial embeddings


rainandbare:
In the `graph.tsx` file - we need to conditionally apply opacity to the canvas element

DoD:

If opacity other than 100% has been applied to canvas dots and we switch to a different mode other than spacial, we should see 100% opaque dots. This should also be true in the side-by-side window.

As per Harley:

"Dot opacity control shouldn’t affect non-spatial embeddings. Currently, changing the dot opacity erroneously decreases the opacity of UMAP and other embeddings (both in side-by-side and switching embedding in the main view)"

[more slack context and images](https://chanzuckerbergteam.slack.com/archives/C06FM5ZQE3Y/p1721159601070449)

@hthomas-czi

PR #1047, opened 2024-07-19
fix: Conditionally apply opacity to the dots

rainandbare:
Fixes issue #1046 

1. Conditionally apply opacity to the canvas element based on `isSpacialMode`

Manual Tests

- [x] Opacity filter effects spacial embedding dots as expected
- [x] When opening a new side-by-side window opacity does not change original window
- [x] When a change is made to the embeddings dropdown and the selection is non-spacial, the dots are always at 100% opacity. 

Let me know if you can think of a e2e test that should be updated.

Screenshot 2024-07-19 at 12 34 53 PM


tihuan (approved):
LGTM 😄 👏 Thanks so much, Suzette! I don't think we have a visual test explicitly for this. Might be a good idea to add one!

rainandbare:
> LGTM 😄 👏 Thanks so much, Suzette! I don't think we have a visual test explicitly for this. Might be a good idea to add one!

Nice! If no objections I am going to merge the change and open a new PR with an opacity controls test.

tihuan:
SGTM! Merging time!! ⏰ ⏰ ⏰

PR #1052, opened 2024-07-24
chore: remove .happy directory and happy-related GH Actions

hspitzley-czi:
- Removes `.happy` directory
- Removes happy build/deploy GH Actions
- Removes Happy-related env vars from Docker
  - Replaces the sha env var with a functional equivalent
  - DEPLOYMENT_STAGE is already specified in `.infra` values files
  - Remove other env vars without replacement , as they were not used in code


jakeyheath (approved):

tihuan (approved):
Thanks so much for the PR!!

PR #1054, opened 2024-07-24
fix: 1053 Truncate histogram and dataset info panel

tihuan:
RDEV: https://absolute-sheepdog.dev-sc.dev.czi.team/d/super-cool-spatial.cxg?spatial=true

<details open>
  <summary>Argus Stack Details</summary>
  <br />
  <table>
    <tr>
      <th>Name</th>
      <td>thuang-1053-add-truncation</td>
    </tr>
    <tr>
      <th>Short Name</th>
      <td>absolute-sheepdog</td>
    </tr>
    <tr>
      <th>Base URL</th>
      <td><a href="https://absolute-sheepdog.dev-sc.dev.czi.team" target="_blank">https://absolute-sheepdog.dev-sc.dev.czi.team</a></td>
    </tr>
  </table>
</details>

---

https://github.com/user-attachments/assets/7c9542ae-c57a-4837-8c0f-f183cd226e5b

Screenshot 2024-07-24 at 2 45 48 PM

Screenshot 2024-07-24 at 12 46 58 PM

### DIFF
--- a/client/src/components/brushableHistogram/histogram.tsx
+++ b/client/src/components/brushableHistogram/histogram.tsx
@@ -5,6 +5,9 @@ import * as d3 from "d3";
 import { AxisDomain } from "d3";
 import maybeScientific from "../../util/maybeScientific";
 import clamp from "../../util/clamp";
+import { truncateText } from "./utils";
+
+const TRUNCATION_MAX_WIDTH_PX = 45;
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
 const Histogram = ({
@@ -130,7 +133,22 @@ const Histogram = ({
                 i: number
               ) => string
             )
-        );
+        )
+        .selectAll("text")
+        .each((_, i, nodes) => {
+          const d3Node = d3.select(nodes[i]);
+          const text = d3Node.text();
+          const node = d3Node.node() as SVGTextElement;
+
+          const truncatedText = truncateText(
+            text,
+            node,
+            TRUNCATION_MAX_WIDTH_PX
+          );
+
+          d3Node.text(truncatedText);
+          d3Node.append("title").text(text);
+        });
 
       /* Y AXIS */
       container
--- a/client/src/components/brushableHistogram/utils.ts
+++ b/client/src/components/brushableHistogram/utils.ts
@@ -1,0 +1,41 @@
+export function truncateText(
+  text: string,
+  node: SVGTextElement,
+  maxWidth: number
+) {
+  let textLength = node.getComputedTextLength();
+
+  if (textLength <= maxWidth) return text; // No truncation needed
+
+  // Calculate average character width and estimate visible characters
+  const avgCharWidth = textLength / text.length;
+  const visibleChars = Math.floor(maxWidth / avgCharWidth);
+
+  // Initial truncation guess
+  let truncatedText = `${text.slice(0, visibleChars - 1)}...`;
+  node.textContent = truncatedText;
+  textLength = node.getComputedTextLength();
+
+  // Fine-tune adjustment if the text is too long
+  if (textLength > maxWidth) {
+    while (textLength > maxWidth && truncatedText.length > 3) {
+      truncatedText = `${truncatedText.slice(0, -4)}...`;
+      node.textContent = truncatedText;
+      textLength = node.getComputedTextLength();
+    }
+  } else {
+    // Fine-tune adjustment if the text is too short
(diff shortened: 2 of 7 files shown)


client/src/components/geneExpression/infoPanel/infoPanelGene/index.tsx
```diff
@@ -24,7 +25,7 @@ function GeneInfo(props: Props) {
       symbol={gene ?? undefined}
       description={geneSummary}
       synonyms={geneSynonyms}
-      references={[]}
```
tihuan, line 27: use `EMPTY_ARRAY` to prevent generating a new `[]` at every render

client/src/components/brushableHistogram/histogram.tsx
```diff
@@ -6,6 +6,8 @@ import { AxisDomain } from "d3";
 import maybeScientific from "../../util/maybeScientific";
 import clamp from "../../util/clamp";
 
+const TRUNCATION_MAX_WIDTH_PX = 45;
```
tihuan, line 10: eye balled the 45px 👀

client/src/components/brushableHistogram/histogram.tsx
```diff
@@ -130,7 +132,31 @@ const Histogram = ({
...
+          const d3Node = d3.select(d3Nodes[i]);
+          const text = d3Node.text();
+          const node = d3Node.node() as SVGTextElement;
+          let textLength = node.getComputedTextLength?.();
+          let truncatedText = text;
+
+          /**
+           * (thuang): Truncate the text if it's too long to fit in the histogram.
+           */
+          if (textLength > TRUNCATION_MAX_WIDTH_PX) {
```
tihuan: it will decrease the text length one char at a time until it fits into the max width. The runtime is horrible, but we're operating on 6 to 8 strings on average, so probably okay.

If optimization is needed, maybe binary could help!

client/src/components/brushableHistogram/utils.ts
```diff
@@ -0,0 +1,41 @@
+export function truncateText(
```
kaloster, line 1: what are your thoughts on this approach - with binary search and the `getComputedTextLength` is extracted to a helper function. We can also potentially memoize `measureText` value but not sure it is necessary here

```
export function truncateText(
  text: string,
  node: SVGTextElement,
  maxWidth: number
) {
  // Helper function to measure text length
  const measureText = (str: string) => {
    node.textContent = str;
    return node.getComputedTextLength();
  };

  // Initial check if truncation is needed
  if (measureText(text) <= maxWidth) return text;

  // Binary search for the truncation point
  let start = 0;
  let end = text.length - 1;
  let truncatedText = "";

  while (start <= end) {
    const mid = Math.floor((start + end) / 2);
    const candidate = `${text.slice(0, mid)}...`;
    const candidateLength = measureText(candidate);

    if (candidateLength <= maxWidth) {
      truncatedText = candidate;
      start = mid + 1;
    } else {
      end = mid - 1;
    }
  }

  return truncatedText;
}

```

client/src/components/brushableHistogram/histogram.tsx
rainandbare: I wonder if there is something sophisticated that could be done with css `text-overflow: ellipsis;` https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/. This seems fine for now though!

rainandbare (approved):

client/src/components/brushableHistogram/utils.ts
tihuan: Great callout! Yeah I actually started with binary search and then ChatGPT suggested that using the text length average as the first guess should get us closer to the target length faster (since it should save a first few binary search iterations that are far from the target)

But also what @rainandbare said, maybe I can just just CSS truncate and text-overflow style to achieve this without hogging the JS thread 😆 

Let me try that real quick!

Thanks both!

client/src/components/brushableHistogram/histogram.tsx
tihuan: Ah dang, okay. it looks like svg text element cannot be styled with width 😭 

https://stackoverflow.com/questions/15975440/add-ellipses-to-overflowing-text-in-svg

So we'll need to use the JS solution still!

client/src/components/brushableHistogram/utils.ts
tihuan: follow up 😭 

https://github.com/chanzuckerberg/single-cell-explorer/pull/1054#discussion_r1692037751

kaloster (approved):

PR #1078, opened 2024-08-21
feat: Add og:image and og:title

sethfeingold:
Changes:

1. Placing og title and image in index.html
2. Import new image in module, so webpack will process it
3. Remove Helmet, since search engines typically don't wait for JS to inject the meta tags in HTML at runtime, and this approach didn't pass the open graph tester. Instead, just add the meta tags in index.html guarantees the best SEO!

Timmy: Confirmed the og:image now works!

Screenshot 2024-08-22 at 8 20 54 AM


tihuan (approved):
Thanks so much for adding this, @sethfeingold 😄 I just made some minor changes to the PR. LGTM 🚀 !!

client/index.html
```diff
@@ -4,6 +4,8 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>CELL&times;GENE | Explorer</title>
+    <meta property="og:title" content="CELLxGENE | Explorer" />
+    <meta property="og:image" content="static/images/cxg-explorer-og.jpg" />
```
kaloster, line 8: 👍🏽 🙏🏽

tihuan: Thanks so much for the review, @kaloster !! 🙏

PR #1080, opened 2024-08-22
fix: skeleton loader

tihuan:
The center Control Skeleton hasn't been working for a while, because the Layout component never called the render function!

This PR fixes it by adding a `...restChildren` in Layout and dump them along with the `graphComponent` in the center panel

Screenshot 2024-08-22 at 1 43 17 PM

https://github.com/user-attachments/assets/e6cc999f-307b-4c76-834f-f0edf6a27b56


kaloster (approved):
🚀 🚀 🚀

PR #1086, opened 2024-08-27
test: SCE-6 e2e tests tabilize snapshots

tihuan:
1. Parallelize tests (22mins to 8mins 🔥 credit: @rainandbare 🏆 ✨ )
1. Extract e2e tests into its own Github action yml file, since Chromatic tests should only be run on `pull` and NOT on `pull_request` for screenshot stability
2. Each Chromatic screenshot requires a UNIQUE snapshot name across the test suite. And this also ensures the screenshots work in `tryUntil()` block, where one or more of the same screenshot could be taken in the tryUntil block because of the retry. Additionally, a explicitly named screenshot promotes intentional screenshot and easier to debug in Chromatic UI page
4. Move `goToPage()` to `cellxgeneActions.ts` to avoid circular dependency
5. Add `resizeWindow()` to ensure the rendered graph is as consistent as possible across different machines. This resolves the problem where locally we could see one sets of lasso counts and in GHA we see another
6. Add `showImageUnderlayInTestMode()` to show the image underlay for snapshots that require the image underlay. This is needed because by default we no longer shows the image underlay, since it introduces instability in Chromatic snapshots if we include it in all the snapshots
7. Add `getSnapshotPrefix(testInfo)` to create a unique snapshot name based on the test name
8. We now crop the graph and image underlay before taking snapshots to reduce the area of image that could cause false positive diff
9. Remove snapshots on tests that don't need them
10. Increase Chromatic `diffThreshold` from the default 6.3% to 10%, since the default was still too sensitive for our image underlay 
11. Test mode now doesn't show the image underlay by default to reduce Chromatic snapshot false positive rate

NOTE:
1. I couldn't completely eliminate the Chromatic snapshot false positives, because technically Chromatic is correct that our image underlay and image are not 100% in sync all the time. So we'll still see 3 to 9 false positives that look like the following and we can just accept them and move on for now:
  Screenshot 2024-09-10 at 12 04 46 PM


client/__tests__/e2e/e2e.test.ts
```diff
@@ -1722,7 +1879,8 @@ test("categories and values from dataset appear and properly truncate if applica
         .all();
 
       expect(Object.keys(categoryRows).length).toBe(1001);
-      await snapshotTestGraph(page, testInfo);
+
+      await takeSnapshot(page, getSnapshotPrefix(testInfo), testInfo);
```
rainandbare, line 1883: Why takeSnapshot and not snapshotTestGraph here?

client/__tests__/e2e/e2e.test.ts
```diff
@@ -1620,6 +1763,7 @@ for (const testDataset of testDatasets) {
 
           test("with continuous legend", async ({ page }, testInfo) => {
             await goToPage(page, url);
+            await showImageUnderlayInTestMode(page);
```
rainandbare, line 1766: should this be in snapshotTestGraph?

client/__tests__/e2e/e2e.test.ts
rainandbare: Oh I see you comment above - maybe different function names or a comment?

client/__tests__/e2e/e2e.test.ts
```diff
@@ -1190,12 +1285,11 @@ for (const testDataset of testDatasets) {
                   }
                 );
 
-                genesHTML = await page
-                  .getByTestId("gene-set-genes")
-                  .innerHTML();
-
-                expect(genesHTML).toMatchSnapshot();
```
rainandbare, line 1197: pesky

client/src/components/graph/graph.tsx
```diff
@@ -922,6 +924,21 @@ class Graph extends React.Component<GraphProps, GraphState> {
       offsetY = (offscreenCanvas.height - targetHeight) / 2;
     }
 
+    /**
+     * (thuang): In test mode, we only want to show a bit of the image underlay
+     * to prevent Chromatic snapshots from creating false positives.
```
rainandbare, line 929: do you wanna include cropping in the comment here?

rainandbare (approved):
This is awesome work @tihuan!!! 🎉 No blocking changes - just some questions.  WOWOWOW!

tihuan (commented):
Oh gosh sorry I forgot to submit my comments 💀 !!

client/__tests__/e2e/e2e.test.ts
tihuan: Great question! There are cases we want to take a snapshot of the graph itself but not the image underlay, so that's why the two functions are separated 🙆‍♂️ Hope that addresses your question!

client/__tests__/e2e/e2e.test.ts
tihuan: Great idea! Adding the comment below thank you 😄 

```ts
/**
   * (thuang): Regular snapshot instead of `snapshotTestGraph`, since this test
   * doesn't involve the graph
   */
  await takeSnapshot(page, getSnapshotPrefix(testInfo), testInfo);
```

client/src/components/graph/graph.tsx
tihuan: Great idea 🙏 

Updated to this:

```ts
/**
     * (thuang): In test mode, we crop the image underlay and the graph
     * to prevent Chromatic snapshots from creating false positives.
     */
```

Thank you!

kaloster (approved):
This is huge @tihuan! Thanks so much for doing this important and thorough work 🙌🏽 

and thanks @rainandbare for parallelizing tests! 🚀

tihuan:
Thanks so much everyone 😄 👏 🎉 !! GO TEAM!! Merginggg

PR #1087, opened 2024-08-28
fix: category label

tihuan:
This should add the label with its associated input correctly!

1. Use Blueprint Checkbox component for input instead of the old Blueprint DOM structure
2. Add `eslint-disable-next-line jsx-a11y/label-has-associated-control` for the whole category label, because it's violating a11y due to using one label for both checkbox and expand/collapse the whole category. So we need UX involved for a complete solution 👌 

No change in UI: 
Screenshot 2024-08-28 at 8 00 37 AM

Bonus: Clicking on a category item will now toggle the checkbox too, so it's easier for users to select and deselect now 🎉 

https://github.com/user-attachments/assets/7bcff2de-1873-4597-8a27-6e16a6449dac


tihuan:
Oh wait let me fix the selection. It's a little wonky at the moment lol

kaloster (approved):
Thanks! I've tried many combinations before, obviously not this one

client/src/components/categorical/category/index.tsx
```diff
@@ -450,6 +450,7 @@ const CategoryHeader = React.memo(
             alignItems: "flex-start",
           }}
         >
+          {/* eslint-disable-next-line jsx-a11y/label-has-associated-control -- TODO: Need to separate expand and checkbox label */}
```
tihuan, line 453: Add `eslint-disable-next-line jsx-a11y/label-has-associated-control` for the whole category label, because it's violating a11y due to using one label for both checkbox and expand/collapse the whole category. So we need UX involved for a complete solution 👌

tihuan:
Ah I need to update snapshot since the DOM structure changed!

client/src/components/categorical/category/index.tsx
kaloster: All of my attempts to treat this error were affecting UI - that's why I opted just to mute it, for now.

Thanks for tackling this and further investigating

PR #1100, opened 2024-09-11
chore: Python 3.12 (SCE-16)

kaloster:
- Updated Python version for GHA workflow
- Updated flatbuffers to 2.0.7
- Updated to Scipy 1.14.1 
- Updated unit tests and related code

### DIFF
--- a/dev_docs/developer_scripts.md
+++ b/dev_docs/developer_scripts.md
@@ -55,8 +55,8 @@ Installs requirements files.
 
 **Usage:** from the `$PROJECT_ROOT/client` directory run `make start-frontend`
 
-NB: the frontend server reads in the desired base_url and dataset name to form the complete url base for API calls. _In
-order to use an arbitrary dataset successfully, the frontend server must be started **after** the backend server_, which
+NB: the frontend server reads in the desired base_url and dataset name to form the complete url base for API calls. \_In
+order to use an arbitrary dataset successfully, the frontend server must be started **after** the backend server\*, which
 writes out the given base_url and dataset anew each time.
 
 #### backend_dev
@@ -67,7 +67,7 @@ the FE developer gets the current version of the backend with a single command
 and no knowledge of python necessary. It creates and activates a virtual
 environment and installs explorer requirements from the current branch.
 
-**Requires** `Python3.6+`, `virtual-env`, `pip`
+**Requires** `Python3.10+`, `virtual-env`, `pip`
 
 **Usage:** from the `$PROJECT_ROOT` directory run `./scripts/backend_dev`
 
@@ -83,7 +83,7 @@ environment and installs explorer requirements from the current branch.
 
 | command                              | purpose                                           |
 | ------------------------------------ | ------------------------------------------------- |
-| python3.11 -m venv explorer          | creates explorer virtual environment              |
+| python3.12 -m venv explorer          | creates explorer virtual environment              |
 | source explorer/bin/activate         | activates virtual environment                     |
 | ./launch_dev_server.sh [cli options] | launches api server (can supply arbitrary config) |
 
--- a/server/common/diffexpdu/postingslist.py
+++ b/server/common/diffexpdu/postingslist.py
@@ -340,7 +340,9 @@ def _un_byteshuffle(buf: Union[bytes, bytearray, memoryview], dtype) -> np.ndarr
 
 
 def _delta(arr: np.ndarray) -> np.ndarray:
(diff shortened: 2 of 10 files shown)


server/tests/unit/dataset/test_cxg_dataset.py
```diff
@@ -51,7 +51,11 @@ def test_tdb_bug(self):
                 )
                 dat = data.summarize_var("mean", filt, 0)
                 summary = decode_fbs.decode_matrix_FBS(dat)
-                self.assertDictContainsSubset({"n_rows": 2638, "n_cols": 1, "row_idx": None}, summary)
```
kaloster, line 54: deprecated

server/common/fbs/matrix.py
```diff
@@ -72,7 +72,7 @@ def encode_matrix_fbs(matrix, row_idx=None, col_idx=None, num_bins=None):
         raise ValueError("FBS Matrix must be 2D")
 
     if sparse.issparse(matrix):
-        matrix = matrix.A
```
kaloster, line 75: AttributeError: 'csc_matrix' object has no attribute 'A'

server/common/diffexpdu/postingslist.py
```diff
@@ -340,7 +340,9 @@ def _un_byteshuffle(buf: Union[bytes, bytearray, memoryview], dtype) -> np.ndarr
 
 
 def _delta(arr: np.ndarray) -> np.ndarray:
-    return np.diff(arr, prepend=arr.dtype.type(0))
```
kaloster, line 343: before this change tests such as `def test_roundtrip()` would fail for precision 0.999

dev_docs/developer_scripts.md
```diff
@@ -55,8 +55,8 @@ Installs requirements files.
 
 **Usage:** from the `$PROJECT_ROOT/client` directory run `make start-frontend`
 
-NB: the frontend server reads in the desired base_url and dataset name to form the complete url base for API calls. _In
-order to use an arbitrary dataset successfully, the frontend server must be started **after** the backend server_, which
+NB: the frontend server reads in the desired base*url and dataset name to form the complete url base for API calls. \_In
```
atarashansky, line 58: why was base_url changed to base*url ?

server/requirements.txt
```diff
@@ -8,7 +8,7 @@ flask-compress==1.13
 flask-cors==5.0.0
 flask-restful==0.3.9
 flask-talisman==1.0.0
-flatbuffers>=1.11.0,<2.0.0 # cellxgene is not compatible with 2.0.0. Requires migration
+flatbuffers==2.0.7
```
atarashansky, line 11: according to comment, flatbuffer 2.0 requires migration. note that you might not get an error after upgrading to 2.0.7 because flatbuffers is only used as part of the dev cycle to generate the autogenerated flatbuffer code. i would not touch this requirement if possible.

atarashansky (changes requested):

dev_docs/developer_scripts.md
kaloster: I think that was just my auto formatting I only changed the python version on this

server/requirements.txt
kaloster: So that means no Python 3.12 - if so we can scrap this PR and stick to 3.11

kaloster: btw - version 2.0.0 doesn't work with Python 3.11 and throws 

```
TypeError: Builder.EndVector() takes 1 positional argument but 2 were given
```

and in 3.12 it throws an error because it is looking for `imp` module which is dropped in this version of python
on some of the unit tests

2.0.7 already updated to use `importlib` and no changes needed outside the ones in this PR 

Could it be that this comment was made against flatbuffers==2.0.0 which is significantly different than 2.0.7?

Unfortunately, flatbuffers only started logging their changes in [2.0.7](https://github.com/google/flatbuffers/blob/master/CHANGELOG.md)

atarashansky: Can we add back a comment to 2.0.7 letting future devs know that the flatbuffer generation has not been tested on any version > 2.0.0 so some debugging might be required?

atarashansky (approved):
https://github.com/chanzuckerberg/single-cell-explorer/pull/1100#discussion_r1755695257

^ just this but otherwise lgtm

server/requirements.txt
kaloster: Yes, great idea! Thanks @atarashansky 🙏🏽

PR #1107, opened 2024-09-18
chore: Add aliases and shared folders for renaming

rainandbare:
This is the beginning of the changes for the Fractal File Structure

In this pr we:
- Add a `common/components` folder for high level shared components
- Add and use aliases for commonly used high level files in order to avoid long file paths as our structure becomes more nested (ie ../../../../../../../../)
- Make aliases compatible with ESLint and Typescript
- Move BottomBanner and BrushableHistogram into new locations


tihuan (approved):
Looks so clean and easy to read!! Legit tears in my eyes 🥹 ✨  

Thank you so much for doing this 🤩 !! 

Should we add this to create-sds-app too as a separate ticket? 🚀🚀🚀

rainandbare:
> Looks so clean and easy to read!! Legit tears in my eyes 🥹 ✨
> 
> Thank you so much for doing this 🤩 !!
> 
> Should we add this to create-sds-app too as a separate ticket? 🚀🚀🚀

Yay! Let's do it! I can get to it early next week!

kaloster (approved):

PR #1117, opened 2024-09-23
chore: bump chart versions

hspitzley-czi:
Updating the chart dependency since it's pretty old at this point. This will update it for staging and rdev, if staging looks good after this is merged I'll make another PR for prod


tihuan (approved):

tihuan:
Thanks so much for upgrading this!

PR #1183, opened 2025-04-25
feat: update url routing for w dataroot

atarashansky:
This is to enable accessing CXGs with arbitrary nesting on the `/w/` dataroot like so: `/user1_data/cxgs/{container_name}.cxg/`


server/app/app.py
```diff
@@ -197,20 +197,28 @@ def __init__(self, app_config: AppConfig):
...
-                f"/{url_dataroot}/{CELLGUIDE_CXG_KEY_NAME}/<path:dataset>.cxg/",
-                f"dataset_index_{url_dataroot}_cellguide_cxgs/",
-                lambda dataset, url_dataroot=url_dataroot: dataset_index(
-                    url_dataroot, f"{CELLGUIDE_CXG_KEY_NAME}/{dataset}.cxg"
-                ),
-                methods=["GET"],
-            )
+            if url_dataroot == "w":
+                self.app.add_url_rule(
+                    f"/{url_dataroot}/<path:dataset>/",
```
kaloster, line 202: Is there anything preventing us from serving this rule in all dataroots - i.e accepting path instead of string and avoid the conditional?

kaloster: or we could potentially extract path/string to a var in the if statement and then plug it into the rule which is otherwise identical

kaloster (approved):
LGTM! just a small improvement suggestion

PR #1189, opened 2025-05-07
feat: Add Cytoband to Chromatin Viz

rainandbare:
First step for this ticket: https://czi.atlassian.net/browse/VC-2528

Next step will be to connect it to the component state to read chromosomeId and svgwidth.
Some tidy up will need to be done with margins and etc but I wanted to get this piece into main as is if possible

Screenshot 2025-05-07 at 1 53 16 PM


client/src/components/CoveragePlot/components/Cytoband/Cytoband.tsx
```diff
@@ -0,0 +1,111 @@
+import * as d3 from "d3";
+import React, { useEffect } from "react";
+import { CytobandWrapper } from "./style";
+import { mockCytobandData, Band } from "./mockCytobandData";
```
rainandbare, line 4: mockCytobandData will come from the api. That can be in the next steps too.

PR #1191, opened 2025-05-08
feat: multiome coverage plot setup

codemonkey800:
[VC-2573](https://czi.atlassian.net/browse/VC-2573)

- Adds chromatin view button to top menu bar
- Updates styling for chromatin view buttons
- Adds minimize button
- Adds `react-query` for fetching coverage plot data
- Integrates coverage plot component with backend via `useCoverageQuery()` hook

### Toggle

image

### Active Toggle

image

### Coverage Plot Buttons

image

### Minimize chromatin view

https://github.com/user-attachments/assets/a2acf588-5cb8-443c-9a12-11abb595bf9e

[VC-2573]: https://czi.atlassian.net/browse/VC-2573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.infra/rdev/values.yaml
+++ b/.infra/rdev/values.yaml
@@ -2,7 +2,7 @@ stack:
   services:
     explorer:
       image:
-        tag: sha-bc0bbac
+        tag: sha-87143f7
       replicaCount: 1
       env:
         # env vars common to all deployment stages


client/src/components/icon/iconSvgPaths.ts
```diff
@@ -13,22 +14,36 @@ interface SvgPathAttribute {
  * Path attributes specs for every supported icon.
  */
 export const IconSvgPaths: Record<IconName, SvgPathAttribute[]> = {
-  about: [
+  [IconNames.ABOUT]: [
```
rainandbare, line 17: Nice!

client/src/hooks/useCoverageQuery.ts
```diff
@@ -0,0 +1,40 @@
+import { UndefinedInitialDataOptions, useQuery } from "@tanstack/react-query";
```
rainandbare, line 1: This is awesome thank you so much for adding this.

client/src/components/CoveragePlot/CoveragePlot.tsx
```diff
@@ -122,6 +147,19 @@ export function CoveragePlot() {
     setHistogramTooltipData(null);
   };
 
+  if (coverageQuery.isLoading) {
```
tihuan, line 150: Love the skeleton state thanks for adding it!

client/src/hooks/useCoverageQuery.ts
```diff
@@ -0,0 +1,40 @@
...
+
+type CoveragePlotData = [number, number, number, number, number][];
```
tihuan, line 3: Can we add a comment to describe what these numbers are 😁? Thank you!

client/src/hooks/useCoverageQuery.ts
tihuan, line 1: Yes!! Love it thanks for adding this, Jeremy 🤩

Discussion: In CxG Data Portal, we have a dedicated folder for queries like so:

https://github.com/chanzuckerberg/single-cell-data-portal/tree/main/frontend/src/common/queries

So it's easier to find queries related functions. Do we wanna do that here too? 😄

client/src/hooks/useCoverageQuery.ts
```diff
@@ -0,0 +1,40 @@
...
+  chromosome: string;
+  cellType: string;
+  options?: Partial<UndefinedInitialDataOptions<FetchCoverageResponse, Error>>
+}
+
+export function useCoverageQuery(
+  { chromosome, cellType, options }: UseCoverageQueryOptions
+) {
+  return useQuery<FetchCoverageResponse, Error>({
+    queryKey: ['fetch-coverage', chromosome, cellType],
```
tihuan, line 36: Discussion: should we use entity `coverage` as key instead of the actual query name?

This way we'd be able to query by entity and invalidate the cache across all CRUD on `coverage`!

Some notes for context:
https://github.com/chanzuckerberg/single-cell-data-portal/tree/main/frontend/src/common/queries#file-structure

Thank you!

client/src/util/fetch.ts
```diff
@@ -0,0 +1,10 @@
+import * as globals from "../globals";
```
tihuan, line 1: Love the extraction thank you!

rainandbare (commented):
For the minimizing we should work with this other window that gets minimied! We can do that in a different ticket if we like. 

https://github.com/user-attachments/assets/23732706-3a97-4fb2-9b51-ed07fe46d610

tihuan (approved):
LGTM! Thanks so much again for adding this ❤️❤️❤️ Just some non-blocking discussions 🤩

client/src/hooks/useCoverageQuery.ts
```diff
@@ -0,0 +1,40 @@
...
+  coveragePlot: CoveragePlotData;
+}
+
+async function fetchCoverage(chromosome: string, cellType: string): Promise<FetchCoverageResponse> {
+  const params = new URLSearchParams();
+  params.set("chr", chromosome);
+  params.set("cell_type", cellType);
+
+  const response = await fetch(
+    `/atac/coverage?${params.toString()}`
```
rainandbare, line 16: One more! We will have to configure this so that it goes to the right url. For local I am finding that the api call works with this url: 

    `http://localhost:5005/s3_uri/example-dataset%252Fsuper-cool-spatial.cxg/api/v0.3/atac/coverage?${params.toString()}`
    
    But I haven't figured out the best way to parameterize it yet.

client/src/hooks/useCoverageQuery.ts
rainandbare: lol - I think we changed it to be [y, startBP, endBP][]

rainandbare: But I am currently working on making the coverage plot show the actual data so I think you can just add a TODO comment here for now.

rainandbare: I think for now that is fine!

I am making changes in this area right now too - I think I have to add the first number back onto the array (the one that is just the index) because the Histogram svg relies on it to render the chart.

PR #1199, opened 2025-05-14
feat: add gene selector and skeleton gene map

rainandbare:


server/dataset/cxg_dataset.py
```diff
@@ -407,14 +407,13 @@ def get_atac_coverage(self, gene_name, genome_version, cell_type):
...
+        target_chromosome, sorted_genes, range_start, range_end = self.get_atac_gene_info(gene_name, genome_version)
 
         # get coverage data (currently random mock data)
-        region_length = 248956422
-        bin_size = 10_000
+        bin_size = 100
         coverage_plot = []
-        for start in range(0, region_length, bin_size):
-            end = min(start + bin_size, region_length)
+        for start in range(range_start, range_end, bin_size):
```
rainandbare, line 415: @kaloster I made these changes so that I could rely on the mock api for bin size and range. Do these work for you?

client/src/components/BottomPanel/components/GeneSelect/GeneSelect.tsx
```diff
@@ -0,0 +1,153 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
+const usePrevious = (value: any) => {
+  const ref = useRef();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
+
+export function GeneSelect() {
```
kaloster, line 26: Probably super low priority at the moment but would be nice if this could be refactored to a more generic component (at some point) as it is duplicating `QuickGenes`

client/src/common/queries/useGenomeReferenceInfo.ts
```diff
@@ -0,0 +1,57 @@
...
+import { fetchJson } from "util/fetch";
+// import { ENTITIES } from "./entites";
+
+export interface FetchGeneInfoResponse {
+  selectedGene: string;
+  chromosomeId: string;
+  allGenesOnChromosome: GeneInfo[];
+}
+
+export interface GeneInfo {
```
kaloster, line 11: Super nit but this is already defined in `common/queries/coverage`

kaloster (approved):
LGTM! 🚀 ⭐

PR #1202, opened 2025-05-15
chore: Remove bottom banner

kaloster:
## Reason for Change

- As per @niknak33 - remove bottom banner from pages


tihuan (approved):
LGTM! Thanks so much for getting a PR out quickly, Ronen 🤩 🙏 ❤️

PR #1207, opened 2025-05-23
feat: Connect cytoband info

rainandbare:


client/src/common/queries/useChromatinViewerSelectedGene.ts
```diff
@@ -9,8 +9,11 @@ export function useChromatinViewerSelectedGene() {
 
   const selectedGeneQuery = useQuery({
     queryKey: SELECTED_GENE_QUERY_KEY,
-    queryFn: () => defaultGene, // return default value
+    queryFn: () => {
```
tihuan, line 12: Quick question: Is `useChromatinViewerSelectedGene()` only used for client state and not making any API calls 😄 ?

If so, I think we should `useState` or `context` to manage it, since ReactQuery should only be used for server state!

client/src/components/BottomPanel/components/ChromosomeMap/ChromosomeMap.tsx
```diff
@@ -1,19 +1,19 @@
...
 import { CoveragePlot } from "../CoveragePlot/CoveragePlot";
 import { GeneMap } from "../GeneMap/GeneMap";
 import { Cytoband } from "../Cytoband/Cytoband";
 import { CoverageAtScale } from "./style";
 
 export const ChromosomeMap = () => {
-  const chromosome = "chr2";
-  const BAR_WIDTH = 8;
-  // const { selectedGene } = useChromatinViewerSelectedGene();
+  const BAR_WIDTH = 6;
```
tihuan, line 14: nit: Can we add some comment around how 6 was determined please 😄 🙏 ? Thank you!

client/src/components/BottomPanel/components/ChromosomeMap/ChromosomeMap.tsx
```diff
@@ -109,6 +77,52 @@ export const ChromosomeMap = () => {
...
+        !binSize
+      ) {
+        return;
+      }
+
+      // Calculate the pixel position relative to the start of the coverage data
+      const relativePosition = selectedGeneInfo.geneStart - startBasePair;
+      const pixelPosition = (relativePosition / binSize) * BAR_WIDTH;
+      // Subtracting 20px to place the gene nicely in the viewport
+      const scrollPosition = Math.max(0, pixelPosition - 20);
```
tihuan, line 108: I wonder if we could add an `id` attribute to each gene annotation svg element and just scroll to that element with `scrollIntoView()`?

Somewhere here?
https://github.com/chanzuckerberg/single-cell-explorer/blob/cdb3ccb3375b4882da108affb3b831f4731f8a07/client/src/components/BottomPanel/components/GeneMap/components/GeneMapSVG/GeneMapSVG.tsx#L248

client/src/components/BottomPanel/components/Cytoband/Cytoband.tsx
```diff
@@ -33,8 +33,19 @@ export const Cytoband = ({
     };
   }, [svgWidth]);
 
+  const cytobandData = useCytobandQuery({
+    chromosome: chromosomeId,
+    genomeVersion: "hg38",
+    options: {
+      enabled: true,
+      retry: 3,
```
tihuan, line 41: IIRC these are the default values, so I think we can remove them!

client/src/common/queries/useChromatinViewerSelectedGene.ts
rainandbare: Ah yes! Can do. Thanks!

PR #1219, opened 2025-06-02
fix: ATAC optimization part 2

kaloster:
This is part 2 of this [optimization PR](https://github.com/chanzuckerberg/single-cell-explorer/pull/1218)

- Update typo in example dataset name
- Added cache per static object
- Coverage endpoint can now be queried for multiple cell_types
- Added FE updates to accommodate the revised endpoint 
```
atac/coverage?gene_name=MYC&genome_version=hg38&cell_types=endothelial cell, leukocyte
```
and will return something like:
```
{
  "chromosome": "chr1",
  "coverageByCellType": {
    "endothelial cell": [[0.0, 1000000.0, 1000500.0], ...],
    "leukocyte": [[0.1, 1000000.0, 1000500.0], ...]
  },
  "geneInfo": [...]
}
```

### DIFF
--- a/.infra/rdev/values.yaml
+++ b/.infra/rdev/values.yaml
@@ -2,7 +2,7 @@ stack:
   services:
     explorer:
       image:
-        tag: sha-0b0ec44
+        tag: sha-739c606
       replicaCount: 1
       env:
         # env vars common to all deployment stages


server/common/cache/atac_cache.py
```diff
@@ -7,6 +8,7 @@
 CYTOBAND_DATA_CACHE = {}
 
 
+@lru_cache(maxsize=4)
```
tihuan, line 11: For my own education, why 4 😄? The default seems to be 128. Is it to make sure we don't blow the memory? Thank you!

tihuan (approved):
LGTM! Thanks so much for speeding up the API, Ronen 🚀 🪩😎

server/common/cache/atac_cache.py
tihuan: Really cool! I see, so the LRU cache is shared between gene and cytoband 💡 That's good to know thank you!

tihuan: Oh wait, Claude could be wrong, but it says each decorator `@lru_cache(maxsize=4)` gets its own LRU cache instance, so if the intention is 2 per entity, then maybe `@lru_cache(maxsize=2)`?

kaloster: I don't have Claude... 😛 but you are correct.

kaloster (commented):
Thank you @rainandbare for fixing the histogram display order!! ⭐

kaloster (commented):
Thank you @rainandbare for fixing the histogram display order!! ⭐

PR #1221, opened 2025-06-04
fix: Style fixes

rainandbare:


client/src/common/components/Embedding/components/ChromatinViewerButton/ChromatinViewerButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const availableCellTypes = cellTypesQuery.data;
+
+    if (availableCellTypes && availableCellTypes.length > 0) {
+      const defaultCellType = availableCellTypes[0];
+      dispatch({
+        type: "toggle chromatin cell types",
+        cellType: defaultCellType,
+      });
+    }
+  }, [cellTypesQuery.data, dispatch]);
```
tihuan, line 63: discussion: For separation of concerns, this feels like `<BottomPanel />` related business logic that the `ChromatinViewerButton` doesn't need to know 😄 ?

Would it make sense to move it to `BottomPanel.tsx` directly?

client/src/common/components/Embedding/components/ChromatinViewerButton/ChromatinViewerButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const chromatinData = useChromatinViewerData(schema, selectedCellTypes);
+  const chromatinButtonDisabled = !chromatinData.data;
+
+  const shouldShowChromatinButton =
+    !isSidePanel &&
+    getFeatureFlag(FEATURES.MULTIOME_VIZ) &&
+    chromatinData.shouldShow;
+
+  useEffect(() => {
+    if (shouldShowChromatinButton && chromatinButtonDisabled) {
```
tihuan, line 78: just wanted to double check on the logic here, do we want to dispatch on `shouldShowChromatinButton && chromatinButtonDisabled` or `shouldShowChromatinButton && !chromatinButtonDisabled`? 

My first reaction is that it'd make sense to open the panel IF `chromatinButtonDisabled` is `false`, indicating that data is available?

Not super sure here 😆

client/src/common/hooks/useChromatinViewerData.ts
```diff
@@ -0,0 +1,87 @@
+import { useEffect } from "react";
+import { useCoverageQuery } from "common/queries/coverage";
+import { Schema } from "common/types/schema";
+import {
+  useChromatinViewerSelectedGene,
+  useChromatinViewerGenome,
+} from "./useChromatinViewerSelectedGene";
+
+// Custom hook to manage chromatin viewer data fetching
+export function useChromatinViewerData(
```
tihuan, line 10: Great hook for consolidating business logics!

client/src/components/BottomPanel/components/CoveragePlot/utils.ts
```diff
@@ -0,0 +1,30 @@
+import memoize from "memoize-one";
+import { AccessionsData, TooltipData } from "./types";
+
+export const generateCoverageVizData = (
+  coverageData: AccessionsData["coverage"],
+  coverageBinSize: number
+) =>
+  coverageData.map((valueArr) => ({
+    x0: valueArr[0] * coverageBinSize - 1, // This is the start of the bin, minus 1 for correct alignment.
```
tihuan, line 9: thanks for the comments!

tihuan (commented):
Looks awesome! Just some minor comments otherwise LGTM 😄 

Please wait for Jeremy's review too 🙏 Thanks so much!

client/src/common/components/Embedding/components/ChromatinViewerButton/ChromatinViewerButton.tsx
rainandbare: I agree that the button is a weird place for that logic!

We need to set the default Cell type before we make the api call to find out if there is any atac coverage data before we know whether we should show the button. So maybe there is a place on load of the app or something? The Bottom Panel isn't loaded yet so we can put it there.

PR #1225, opened 2025-06-13
fix: No duplicate cell types tracks

rainandbare:
Fixed issue where you were able to have both tracks be the same cell type which caused logic issues.


kaloster (approved):
LGTM! 🚀 🎆 🎉

PR #1230, opened 2025-06-24
feat: Updates to Gene panel readability

rainandbare:
https://driven-ant.dev-sc.dev.czi.team/d/super-cool-atac.cxg/?multiome_viz=true

This is to update the gene panel with a drop down menu and to include the Open Multiome Viz in the side bar.


kaloster (approved):
Cool! 🔥

Issue #1233, opened 2025-07-09
[QUESTION]


alivernois:
Hello all!

I tried searching the Tabula Sapiens cell atlas (and others) for reference genes (for eg, GAPDH), but they aren't found in the datasets. I've done this in the recent past and the genes are there. Does anyone know what's changed?

[tabula_sapiens.pdf](https://github.com/user-attachments/files/21149913/tabula_sapiens.pdf)

MaximilianLombardo:
Hi @alivernois - thanks for reporting this - our team has been looking into this and are prioritizing working on a fix - hang on tight!

Issue #1231, opened 2025-07-08
[BUG] CZ CELLxGENE Discover platform fails to access gene data.


gennadyFauna:
**Describe the bug**
The main platform, hosted at https://cellxgene.cziscience.com, recently broke in several ways. Please do tell me if this should be reported elsewhere. So far, I have noticed two key issues:

- Differential expression fails to return results.
- The platform fails to display genes present in data, and appears to be biased toward miscellaneous RNA species. 

**To Reproduce**
This example is for a single dataset I happen to have on hand, but I see the same issues in spot checks in other datasets. 

Steps to reproduce the DE bug:
1. Go to [the HYPOMAP dataset](https://cellxgene.cziscience.com/collections/d0941303-7ce3-4422-9249-cf31eb98c480). Open the dataset using "Explore".
2. Select any two distinct populations of cells.
3. Press "see top differentially expressed genes."
4. The outputs under "Pop 1 high" and "Pop 2 high" read `No genes to display`.

Steps to reproduce the present genes bug:
1. Download the HYPOMAP dataset. Consider the top genes by total expression, or a standard marker (SNAP25). These genes will not show up.
2. So far as I can tell, this is not a split by `feature_type`. Based on spot checks, some protein-coding genes (OR4F16, HES4, ISG15, etc.) appear on the platform.
3. Based on spot checks, the following features do not appear on platform: `processed_pseudogene`, `snRNA`, `miRNA`, `unprocessed_pseudogene`, `misc_RNA` (although there is a vast number of undifferentiated Y_RNA entries), `transcribed_unprocessed_pseudogene` (although SLC35E2A shows up), `rRNA_pseudogene`, `rRNA`, `Mt_tRNA`.
4. The following features do appear: `lncRNA` (but not all: a vast number of LINCs show up, but not not MALAT1), `snoRNA` (but not all: SNORA72 shows up but SNORD118 does not). I stopped looking here, as these categories cover most of the genes. 

**Expected behavior**

- DE should always work.
- Genes present in the data matrix should always appear.

**Screenshots**
Today's version of the dataset (`f91ae570a42434e8066d0b512584d711`), and for that matter the October one I had on hand (not shown, `dd5ffb49acfe2cb437d2fdee541e62df`), have the correct genes (here, sorted by total expression). 

<img width="961" height="525" alt="Image" src="https://github.com/user-attachments/assets/8e75be84-e2d9-4284-9ece-328d118ebb4f" />

[...]

PR #1237, opened 2025-07-10
fix: Update Explorer to use feature_name

kaloster:
<details open>
  <summary>Argus Stack Details</summary>
  <br />
  <table>
    <tr>
      <th>App Name</th>
      <td>single-cell-explorer</td>
    </tr>
    <tr>
      <th>Stack Name</th>
      <td>natural-hermit</td>
    </tr>
    <tr>
      <th>Base URL</th>
      <td><a href="https://natural-hermit.dev-sc.dev.czi.team" target="_blank">https://natural-hermit.dev-sc.dev.czi.team</a></td>
    </tr>
  </table>
</details>

---

Related to: https://github.com/chanzuckerberg/single-cell-explorer/issues/1231 and https://github.com/chanzuckerberg/single-cell-explorer/issues/1233

- Updated key functionality to use `feature_name` var column to support post 6.0 schema migrations CxGs
- Reverted the previous implementation
- Added fallback to support test dataset and scenarios where `feature_name` column is not present

### DIFF
--- a/client/src/actions/geneset.ts
+++ b/client/src/actions/geneset.ts
@@ -55,9 +55,18 @@ export const genesetAddGenes =
   (genesetName: any, genes: any) => async (dispatch: any, getState: any) => {
     const state = getState();
     const { obsCrossfilter: prevObsCrossfilter, annoMatrix } = state;
+    const { schema } = annoMatrix;
+    // Fallback
+    const varIndex = schema.annotations.var.index;
     const varFeatureName = VAR_FEATURE_NAME_COLUMN;
-    const df: Dataframe = await annoMatrix.fetch("var", varFeatureName);
-    const geneNames = df.col(varFeatureName).asArray();
+    // Check if VAR_FEATURE_NAME_COLUMN exists, otherwise use the index
+    const varColumns = annoMatrix.getMatrixColumns("var");
+    const columnName = varColumns.includes(varFeatureName)
+      ? varFeatureName
+      : varIndex;
+
+    const df: Dataframe = await annoMatrix.fetch("var", columnName);
+    const geneNames = df.col(columnName).asArray();
     // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
     genes = genes.reduce((acc: any, gene: any) => {
       if (geneNames.indexOf(gene.geneSymbol) === -1) {
--- a/client/src/actions/geneset.ts
+++ b/client/src/actions/geneset.ts
@@ -21,6 +21,7 @@ The behavior manifest in these action creators:
 Note that crossfilter indices are lazy created, as needed.
 */
 
+import { VAR_FEATURE_NAME_COLUMN } from "../common/constants";
 import { Dataframe } from "../util/dataframe";
 
 export const genesetDelete =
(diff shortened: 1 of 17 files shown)


client/src/actions/geneset.ts
```diff
@@ -54,16 +55,9 @@ export const genesetAddGenes =
...
-    const varIndex = schema.annotations.var.index;
-    const varLabel = "feature_name";
-    const varColumns = annoMatrix.getMatrixColumns("var");
-
-    const labelToUse = varColumns.includes(varLabel) ? varLabel : varIndex;
-    const dfNames: Dataframe = await annoMatrix.fetch("var", labelToUse);
-
-    const geneNames = dfNames.col(labelToUse).asArray();
-
+    const varFeatureName = VAR_FEATURE_NAME_COLUMN;
```
tihuan, line 58: sanity check - maybe we could just use `VAR_FEATURE_NAME_COLUMN` directly at callsite?

```ts
const geneNames = df.col(VAR_FEATURE_NAME_COLUMN).asArray();
```

client/src/actions/index.ts
```diff
@@ -434,7 +435,8 @@ const requestDifferentialExpression =
 
       const response = await res.json();
 
-      const varIndex = await annoMatrix.fetch(Field.var, varIndexName);
+      const varLabel = await annoMatrix.fetch(Field.var, varFeatureName);
```
tihuan, line 438: Maybe just `const varLabel = await annoMatrix.fetch(Field.var, VAR_FEATURE_NAME_COLUMN);` 😄 ?

client/src/common/components/BrushableHistogram/BrushableHistogram.tsx
```diff
@@ -487,14 +488,16 @@ class HistogramBrush extends React.PureComponent<BrushableHistogramProps> {
     const varIndex = schema?.annotations?.var?.index;
     if (!varIndex) return null;
```
tihuan, line 489: should we remove `if (!varIndex) return null;` since the component doesn't rely on it anymore 😄 ?

client/src/common/constants.ts
```diff
@@ -1 +1,4 @@
 export const EMPTY_ARRAY = [];
+
+// Column name for gene names in the annomatrix
```
tihuan, line 3: Not a blocker for now, but my only concern with this approach is that it's not as flexible as the original design's intention to allow the dataset to decide which column to use as feature anymore?

If we're seeing issues with it, would it be possible that it's the datasets that's passing the wrong feature column as index?

tihuan:
Looks like `feature_name` doesn't exist!

Screenshot 2025-07-10 at 7 41 07 AM

tihuan:
> Looks like `feature_name` doesn't exist!
> 
> Screenshot 2025-07-10 at 7 41 07 AM

Ronen says that the example datasets are failing because they're not on schema 6.0 yet!

kaloster:
@tihuan Just added fallback to accommodate for test datasets

server/default_config.py
```diff
@@ -93,7 +93,7 @@
     about_legal_privacy: null
 
   presentation:
-    max_categories: 30000
+    max_categories: 1000
```
tihuan, line 96: non-blocking: just curious, why is this change needed again thank you!

kaloster: Oh because I changed it yesterday, to see if it would solve a different issue - just reverting to the way it was

PR #1244, opened 2025-07-23
fix: Grab organism_ontology_term_id from config

kaloster:
Updated the ChromatinViewerProvider to dynamically detect genome version and default gene (MYC for human, Vip for mouse) from the dataset's organism ontology term ID.


tihuan (approved):
LGTM thank you for the quick fix!!

PR #1272, opened 2025-09-25
feat: UW-10 explorer byod toast

ehoops-cz:
We want to show a toast on explorer that links to BYOD.  This is currently behind the byod feature flag.

Both the toast and modal were copied from the discover documentation. However, the icon uses the explorer icon.

I'm struggling to get the font to update.  Right now it matches the rest of explorer - @Janeece should I work on updating the font or leave it to match explorer.

Screenshot 2025-09-25 at 2 27 12 PM
Screenshot 2025-09-25 at 2 27 57 PM


Janeece (approved):
We _should_ update the fonts if it's possible, especially for the dialog. 
That said, it's not the top priority IMO — if its gonna take too much time at the cost of the other things to do, we could come back later to try to fix it. 

The toast is looking good aside from the fonts.

ehoops-cz:
> We _should_ update the fonts if it's possible, especially for the dialog. That said, it's not the top priority IMO — if its gonna take too much time at the cost of the other things to do, we could come back later to try to fix it.
> 
> The toast is looking good aside from the fonts.

Sounds good.  I won't block this PR on fixing the fonts, but I'll plan to come back to this once the other BYOD related work is done.

client/src/components/BYODConnectors/BYODToast/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+      .sort((a, b) => b.getTime() - a.getTime()); // Sort descending
+
+    const latestPassedDate = passedForcedDates.length > 0 ? passedForcedDates[0] : null;
+    const lastForcedDatePassed = latestPassedDate ?
+      latestPassedDate.toISOString().split('T')[0] : undefined;
+
+    const state: ToastDismissalState = {
+      toastId,
+      dismissedAt: Date.now(),
+      dismissed: true,
```
egh, line 90: would this value ever be false?

client/src/components/BYODConnectors/BYODToast/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+      lastForcedDatePassed,
```
egh, line 91: I think I'm missing something here because if the last forced date has passed, isn't it enough to store the dismissed at date?

client/src/components/BYODConnectors/BYODToast/utils.ts
```diff
@@ -0,0 +1,101 @@
+const TOAST_STORAGE_PREFIX = "byod-toast-dismissed-";
```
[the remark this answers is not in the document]
ehoops-cz: Updated, thanks! this is much more readable

egh (approved):
👍

